PR #1, opened 2025-03-22
chore(config): fix typos

Crunchyman-ralph:
When running: ```npx claude-task-init``` it doesn't work

We should use the updated ```npx task-master-ai``` instead


eyaltoledano (approved):

eyaltoledano:
Thanks for this. That's what I get for renaming the project before release lol

PR #9, opened 2025-03-24
feat: implement MCP

Crunchyman-ralph:
- Implemented an MCP server, but seems like there is a specific response that cursor and other AI agents expect to be able to use the MCP.
- Might just use a package that handles that for me, need to research some more stuff.

- This PR is a WIP, any comments welcome :D

TODO:
- [ ] Implement standardized jsonrpc response for cursor mcp client to understand
- [ ] Implement tool definitions for mcp to understand and use
- [ ] Create handlers that translate tool calls to CLI commands
- [ ] Implement proper error handling specific to each tool
- [ ] Add tool discovery and introspection capabilities
- [ ] Make it compatible with Cursor (apparently even though MCP is supposed to be standardized, each tool handles it differently, so we need to create adapters for cursor, claude, etc...)

### DIFF
--- a/.cursor/mcp.json
+++ b/.cursor/mcp.json
@@ -1,0 +1,11 @@
+{
+  "servers": [
+    {
+      "name": "task-master",
+      "type": "command",
+      "command": "npx",
+      "args": ["task-master", "mcp"],
+      "description": "Task Master MCP - process task-master commands"
+    }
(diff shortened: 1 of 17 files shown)


Crunchyman-ralph:
Closed in favor of #10

PR #20, opened 2025-03-25
feat: Implement MCP

### LINKED ISSUE #7: Introduce the Taskmaster MCP server
- Right now we function through commands, we should add MCP to make this more AI-friendly

Crunchyman-ralph:
- All works, test the MCP, works great.
- Implemented an mcp server that you can start with npx task-master-mcp, wanted to have a separate npx so that we can have a good separation of concerns for people who might want to use the cli exclusively or mcp exclusively
- Closes #7


chad-fossa:
Consider adding instructions on setting up a .mcp.json file, including possible configuration.  As opposed to just setting this up for Cursor, providing instructions on making a .mcp.json file (or just referencing the one you wrote) can help for setup with Claude Code.

Also, consider providing a Dockerfile and instructions to build the dockerfile (and make the .mcp.json file reference the docker file you built with an associated volume) similar to https://github.com/modelcontextprotocol/servers/tree/main/src/memory

tasks/tasks.json
```diff
@@ -1343,8 +1343,68 @@
         22
       ],
       "priority": "medium",
```
chad-fossa, line 1345: It might be easier to get this PR merged if you only included the "mcp-server" folder, a brief mention in the main README.md file, and a separate README.md inside the mcp-server folder.

package.json
```diff
@@ -6,15 +6,17 @@
...
+    "task-master-mcp-server": "mcp-server/server.js"
   },
   "scripts": {
     "test": "node --experimental-vm-modules node_modules/.bin/jest",
     "test:watch": "node --experimental-vm-modules node_modules/.bin/jest --watch",
     "test:coverage": "node --experimental-vm-modules node_modules/.bin/jest --coverage",
     "prepare-package": "node scripts/prepare-package.js",
     "prepublishOnly": "npm run prepare-package",
-    "prepare": "chmod +x bin/task-master.js bin/task-master-init.js"
+    "prepare": "chmod +x bin/task-master.js bin/task-master-init.js mcp-server/server.js",
```
chad-fossa: [consider] add a "prepare:mcp" command rather than changing the default functionality

mcp-server/src/index.js
```diff
@@ -0,0 +1,86 @@
...
+dotenv.config();
+
+// Constants
+const __filename = fileURLToPath(import.meta.url);
+const __dirname = path.dirname(__filename);
+
+/**
+ * Main MCP server class that integrates with Task Master
+ */
+class TaskMasterMCPServer {
```
chad-fossa, line 19: Probably better to use the actual MCP "Server" class from the sdk: https://github.com/modelcontextprotocol/servers/blob/main/src/memory/index.ts#L3

mcp-server/src/index.js
```diff
@@ -0,0 +1,86 @@
+import { FastMCP } from "fastmcp";
```
chad-fossa, line 1: Note that "fastmcp" is now part of the official "MCP SDK"

chad-fossa (commented):
Consider moving to the MCP SDK, and adding some test coverage.

mcp-server/src/index.js
Crunchyman-ralph: How would that look ? wouldn't that conflict with the use of FastMCP ?
Creating a custom class and using FastMCP inside it.

If I were to use the server from the SDK, wouldn't that stop me from using FastMCP ?

package.json
Crunchyman-ralph: good point, probably don't even need this here, good catch

tasks/tasks.json
Crunchyman-ralph: I wouldn't want to have multiple Readme especially since the MCP will slowly be an integral part of this project

mcp-server/src/index.js
chad-fossa: Pretty sure it would replace fastmcp since fastmcp is now part of the sdk (according to their GitHub)

eyaltoledano: ORLY? so SDK is built on top of fastmcp now?

If so, then I'd assume the SDK further simplifies the setup though tbh fastmcp is REALLY simple and comes with a command to easily install to Claude too fwiw

tasks/tasks.json
eyaltoledano: Either way I just checked out and pulled this branch and merged with main flawlessly with not a single conflict so setting up in the mcp-server folder and nothing else was the right call -- well done

mcp-server/src/index.js
Crunchyman-ralph: Probably won't do for this PR, but will definitely need to do soon, as a refactor.

It works now and this needs to go out ASAP.

README-task-master.md
```diff
@@ -123,6 +126,21 @@ Claude Task Master is designed to work seamlessly with [Cursor AI](https://www.c
...
+
+To enable enhanced task management capabilities directly within Cursor using the Model Control Protocol (MCP):
+
+1. Go to Cursor settings
+2. Navigate to the MCP section
+3. Click on "Add New MCP Server"
+4. Configure with the following details:
+   - Name: "Task Master"
+   - Type: "Command"
+   - Command: "npx -y task-master-mcp"
```
rtuin: Suggestion: when this is published to npm, having the `--package` argument is safer. This prevents that when one creates and publishes an npm package called `task-master-mcp` their package gets executed.

```suggestion
   - Command: "npx -y --package task-master-ai task-master-mcp"
```

README.md
```diff
@@ -123,6 +126,21 @@ Claude Task Master is designed to work seamlessly with [Cursor AI](https://www.c
...
+
+To enable enhanced task management capabilities directly within Cursor using the Model Control Protocol (MCP):
+
+1. Go to Cursor settings
+2. Navigate to the MCP section
+3. Click on "Add New MCP Server"
+4. Configure with the following details:
+   - Name: "Task Master"
+   - Type: "Command"
+   - Command: "npx -y task-master-mcp"
```
rtuin: See https://github.com/eyaltoledano/claude-task-master/pull/20/files#r2017643913

```suggestion
   - Command: "npx -y --package task-master-ai task-master-mcp"
```

README-task-master.md
Crunchyman-ralph: good catch!

PR #50, opened 2025-03-29
fix: addTask mcp tool not working

Crunchyman-ralph:
- Fix `projectRoot not defined` when trying to add task from the mcp


Crunchyman-ralph:
Minor fix, self-merging

Issue #68, opened 2025-03-30
Anyone having issues with MCP set up in Cursor?


Abdo-El-Mobayad:
I'm trying to set up the task master mcp in my current project & getting a failed to create client error message. 
Any ideas how to sort this out?

![Image](https://github.com/user-attachments/assets/803118b2-7020-40de-998f-693675612fda)

eyaltoledano:
The current live MCP implementation is mega janky but the connection to the client should be working.

I see you added it globally. What did you enter into the +Add new global MCP server form?

Crunchyman-ralph:
Yeah I figured, in the next release, that MCP connection should be fixed.

neves8232:
I have a problem where it always states client closed. When trying to run the command directly I get "sh: 1: task-master-mcp: not found" I've installed the global version tho

Abdo-El-Mobayad:
@eyaltoledano 
I set it up like this inside the MCP Json Config file

"Task-Master": {
  "command": "cmd",
  "args": ["/c", "npx", "-y", "--package", "task-master-ai", "task-master-mcp"]
  }
}

Also is it possible to add the Anthropic API directly inside the Cursor Config like this?

"Task-Master": {
  "command": "cmd",
  "args": ["/c", "npx", "-y", "--package", "task-master-ai", "task-master-mcp"],
  "env": {
    "ANTHROPIC_API_KEY": "your_actual_api_key_here"
  }
}

eyaltoledano:
> [@eyaltoledano](https://github.com/eyaltoledano) I set it up like this inside the MCP Json Config file
> 
> "Task-Master": { "command": "cmd", "args": ["/c", "npx", "-y", "--package", "task-master-ai", "task-master-mcp"] } }
> 
> Also is it possible to add the Anthropic API directly inside the Cursor Config like this?
> 
> "Task-Master": { "command": "cmd", "args": ["/c", "npx", "-y", "--package", "task-master-ai", "task-master-mcp"], "env": { "ANTHROPIC_API_KEY": "your_actual_api_key_here" } }

Thanks for bringing this up, I need to add this in.

The actual command should probably be something like `npx -y task-master task-master-mcp-server` but will confirm when live

AbrahamWisman:
I tried with the new 0.10.2 but nothing seems to work for me, when running directly in terminal it cant find the command

lexprimost:
[...]

PR #71, opened 2025-03-31
23.16 23.30

eyaltoledano:
Working through moving the MCP functions to use the direct functions.

MCP is feature complete, the last thing missing is to ensure the root is properly handled.


.changeset/two-bats-smoke.md
Crunchyman-ralph, line 1: Not a patch, this is a minor for sure

Crunchyman-ralph (commented):
Not going to review yet, just a small nit pick

eyaltoledano:
Did i use changeset properly or still need adjutsments

Crunchyman-ralph:
Yup pretty good

eyaltoledano:
OK so the issue with the root path stuff is that we're conflating two paths in one

If we think about it, we have 2 paths to keep track of:

**Package Path:** 
- Where the task-master-ai npm package is installed
- For global installs: In the global node_modules directory
- For local installs: In ./node_modules/task-master

**Project Path:** 
- Where the user's project containing tasks/tasks.json is located
- This is where tasks.json should be found
- Usually the current working directory or a parent directory

So we need to be identify and grabbing both

This is the main blocker for MCP rn

eyaltoledano:
> OK so the issue with the root path stuff is that we're conflating two paths in one
> 
> If we think about it, we have 2 paths to keep track of:
> 
> **Package Path:**
> 
> * Where the task-master-ai npm package is installed
> * For global installs: In the global node_modules directory
> * For local installs: In ./node_modules/task-master
> 
> **Project Path:**
> 
> * Where the user's project containing tasks/tasks.json is located
> * This is where tasks.json should be found
> * Usually the current working directory or a parent directory
> 
> So we need to be identify and grab both in a recursive way. researching
> 
> This is the main blocker for MCP rn

I've updated subtask 23.38 with more specific information about the path handling challenges and advanced solutions.

The update includes:

**Key Challenges to Avoid:**
- Relying solely on process.cwd() with detailed explanation of why it's problematic
- Dual Path Requirements clearly defined (Package Path vs Project Path)
- Specific Edge Cases that need to be handled (non-project directories, nested structures, workspaces, etc.)

**Advanced Solutions:**
- Project Marker Detection using find-up package to search for package.json or .git
- Package Path Resolution using import.meta.url with fileURLToPath
- Workspace-Aware Resolution for Yarn/pnpm workspaces
- Monorepo Handling with cascading configuration search
- CLI Tool Inspiration from ESLint, Jest, and Next.js
- Robust Path Resolution Algorithm with code example
- Environment Variable Overrides for explicit path setting

This additional information should help us do the path handling in task-master in a way that should work across all npm installation scenarios without requiring manual path specification.

The research confirms using import.meta.url for package path and recursive directory traversal for project path, while also suggesting additional techniques like using environment variables and dedicated packages like find-up for even more robust handling.

eyaltoledano:
This is now feature complete and tested on a fresh install with `npm link`

Two main changes
- implemented some sexy path discovery. we now have two paths: the project path, and the package path. both are important for separate reasons
- i adjusted the mcp server command in the mcp.json that we had. this is why it is broken for users. it will fix #68 #50 

I'm seeing some commands still failing to find the project root but they may just need a touch to work list the list-tasks does, which doesnt need the root. Fixing now and pushing here, at which point i only need to check in on all the other tasks to verify no errors, then update the rules and readme, and get this into next for a release

Readme really will need some help

Gonna be a busy day tomorrow

PR #95, opened 2025-04-04
Fix version discrepancy between code, package.json, and GitHub releases

ichoosetoaccept:
# Fix Version Discrepancy Issues

## Problem
There were multiple version inconsistencies in the Task Master CLI:
1. The hardcoded version in utils.js was "1.5.0"
2. The package.json version was "0.9.30"
3. The GitHub release was "v0.10.2"
4. The version detection logic was unreliable, sometimes showing the wrong version

## Solution
This PR fixes these issues by:
1. Removing the hardcoded version in utils.js
2. Updating the package.json version to match the GitHub release (0.10.2)
3. Improving the version detection logic to reliably find and use the correct version
4. Adding better error handling with debug logging

## Testing
Verified that `task-master --version` and the banner in `task-master --help` both correctly show version 0.10.2.

This ensures consistent versioning across the codebase and GitHub releases, providing a clear path forward for future releases.

### DIFF
--- a/bin/task-master.js
+++ b/bin/task-master.js
@@ -19,7 +19,9 @@ const __dirname = dirname(__filename);
 const require = createRequire(import.meta.url);
 
 // Get package information
-const packageJson = require('../package.json');
+// Use a more reliable path to find the package.json
+const packageJsonPath = resolve(__dirname, '../package.json');
+const packageJson = require(packageJsonPath);
 const version = packageJson.version;
 
 // Get paths to script files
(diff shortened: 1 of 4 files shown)


bin/task-master.js
Crunchyman-ralph, line 1: I like this change, think its important

package.json
Crunchyman-ralph, line 1: please remove this, will be fixed separetly, its due to a different issue

Crunchyman-ralph (commented):
I'd like to accept this PR, if you can just clean it up and fix conflicts, I'd love to give it another review

Crunchyman-ralph:
Taking over this PR

Crunchyman-ralph:
Changed my mind, we'll come back to this later, keeping it as is for now

PR #129, opened 2025-04-09
fix: Remove task-master-ai as a dependency from the package.json gene…

eyaltoledano:
Removes the task-master-ai package added during init.js section adding to user's package json
We need to improve this to be closer to convention as we don't want to pollute users' local files


Crunchyman-ralph (approved):
lgtm

PR #167, opened 2025-04-11
wip: refactor(mcp): Enforce projectRoot and centralize path validation

eyaltoledano:
This commit refactors how project paths are handled in MCP direct functions to improve reliability, particularly when session context is incomplete or missing.

Key changes:

1.  **Made `projectRoot` required in MCP tools:**
    *   Updated `analyze_project_complexity` tool (`analyze.js`) Zod schema to require the `projectRoot` parameter.
    *   Removed logic in the tool's `execute` method that attempted to derive `projectRoot` from the session as a fallback.

2.  **Refactored `findTasksJsonPath` utility (`path-utils.js`):**
    *   Modified to return both `{ tasksPath, validatedProjectRoot }`.
    *   Added validation to ensure the determined `projectRoot` exists and is a directory.
    *   Requires the `session` object to be passed (for internal logic, even if root comes from args).

3.  **Updated all Direct Functions:**
    *   Modified signatures to accept `{ session }` from context: `async function fn(args, log, { session })`.
    *   Updated all calls to `findTasksJsonPath` to pass the `session` object: `findTasksJsonPath(args, log, session)`.

4.  **Updated `analyzeTaskComplexityDirect`:**
    *   Uses the `validatedProjectRoot` returned by `findTasksJsonPath` for resolving the output report path via `resolveProjectPath`.
    *   Uses `ensureDirectoryExists` for the output path directory.

This ensures that operations relying on project context receive an explicitly provided and validated project root directory, resolving errors caused by incorrect path resolution (e.g., falling back to '/').

### DIFF
--- a/mcp-server/src/core/direct-functions/add-dependency.js
+++ b/mcp-server/src/core/direct-functions/add-dependency.js
@@ -21,7 +21,7 @@ import {
  * @param {Object} log - Logger object
  * @returns {Promise<Object>} - Result object with success status and data/error information
  */
-export async function addDependencyDirect(args, log) {
+export async function addDependencyDirect(args, log, { session }) {
 	try {
 		log.info(`Adding dependency with args: ${JSON.stringify(args)}`);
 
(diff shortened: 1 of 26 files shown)


eyaltoledano:
Stale

PR #169, opened 2025-04-11
Don't add task-master-mcp to mcp.json if it already exists

joedanz:


scripts/init.js
```diff
@@ -935,6 +938,23 @@ function setupMCPConfiguration(targetDir, projectName) {
 				mcpConfig.mcpServers = {};
 			}
 
+			// Check if any existing server configuration already has task-master-mcp in its args
+			const hasMCPString = Object.values(mcpConfig.mcpServers).some(
+				(server) =>
+					server.args &&
+					server.args.some(
+						(arg) => typeof arg === 'string' && arg.includes('task-master-mcp')
```
Crunchyman-ralph, line 946: the key is called task-master-ai, but we should probably store a variable of the name somewhere global to avoid these types of issues...

Crunchyman-ralph (changes requested):
small change

Crunchyman-ralph (approved):
lgtm

Issue #79, opened 2025-04-01
Better AI Model Management


eyaltoledano:
> "I need more control and visibility over which AI models Task Master uses for different operations, especially as more models become available. Having an easy way to manage and switch models for main operations and research tasks would greatly enhance flexibility and customization."

---

### Motivation
Currently, Task Master only supports Claude for main operations and Perplexity for research. Users are limited in flexibility when managing AI models. Adding comprehensive support for multiple popular AI models (OpenAI, Ollama, Gemini, OpenRouter, Grok) and providing intuitive CLI commands for model management will significantly enhance usability, transparency, and adaptability to user preferences and project-specific needs.

#38 

---

### Proposed Solution
Implement an intuitive CLI command for AI model management:

- `task-master models`: Lists currently configured models for main operations and research.
- `task-master models --set-main="<model_name>" --set-research="<model_name>"`: Sets the desired models for main operations and research tasks respectively.

Supported AI Models:
- **Main Operations:** Claude (current default), OpenAI, Ollama, Gemini, OpenRouter
- **Research Operations:** Perplexity (current default), OpenAI, Ollama, Grok

If a user specifies an invalid model, the CLI lists available models clearly.

### Example CLI Usage

List current models:
```shell
task-master models
```
Output example:
```
Current AI Model Configuration:
- Main Operations: Claude
- Research Operations: Perplexity
```

Set new models:
```shell
task-master models --set-main="gemini" --set-research="grok3"
```

Attempt invalid model:
```shell
task-master models --set-main="invalidModel"
```
Output example:
```
Error: "invalidModel" is not a valid model.

Available models for Main Operations:
- claude
- openai
- ollama
- gemini
- openrouter
```

---

### High-Level Workflow
1. Update CLI parsing logic to handle new `models` command and associated flags.
2. Consolidate all AI calls into `ai-services.js` for centralized management.
3. Implement robust wrapper functions for each AI API:
   - Claude (existing)
   - Perplexity (existing)
   - OpenAI
   - Ollama
   - Gemini
   - OpenRouter
   - Grok
4. Update environment variables and provide clear documentation in `.env_example`:
```env
# MAIN_MODEL options: claude, openai, ollama, gemini, openrouter
MAIN_MODEL=claude

# RESEARCH_MODEL options: perplexity, openai, ollama, grok
RESEARCH_MODEL=perplexity
```
5. Ensure dynamic model switching via environment variables or configuration management.
[...]

PR #172, opened 2025-04-11
chore(ai): Reduces context window back from 128k to 64k

eyaltoledano:
Until we decouple context windows between main and research models per #79


Crunchyman-ralph (dismissed):

Crunchyman-ralph (approved):
lgtm

PR #183, opened 2025-04-12
Update README.md

sleeper:
Fix MCP Server configuration for Cursor


eyaltoledano:
That's no longer the command we use. The command in the readme is correct.

sleeper:
OK, but I tried the command in the README, and it did not work. Henc,e the proposal.

PR #190, opened 2025-04-12
fix: README bug not showing precise instructions

Crunchyman-ralph:
- Users who don't have the package `task-master-ai` installed, can't use npx -y task-master-mcp...


eyaltoledano (approved):
lgtm

Issue #184, opened 2025-04-12
bug:


murphman300:
### Description

Detailed description of the problem, including steps to reproduce the issue.

### Steps to Reproduce

1. Step-by-step instructions to reproduce the issue
2. Include command examples or UI interactions

`task-master init` after running npm I - g task-master-ai

### Expected Behavior

Describe clearly what the expected outcome or behavior should be.

### Actual Behavior

Describe clearly what the actual outcome or behavior is.

running task-master init does not work

### Screenshots or Logs

Provide screenshots, logs, or error messages if applicable.

### Environment

- Task Master version: latest
- Node.js version: 20.14
- Operating system: ubuntu 22 LTS on WSL 2
- IDE (if applicable): cursor

### Additional Context

Any additional information or context that might help diagnose the issue.
Getting this from task-master init

task-master init
/usr/bin/env: ‘node --trace-deprecation’: No such file or directory
/usr/bin/env: use -[v]S to pass options in shebang lines

arcangelo7:
Same bug here

Crunchyman-ralph:
checking it out, thanks for the PR

eyaltoledano:
> Added a PR fix for it. In the meantime, I manually edited the file:
> 
> ```
> # open task-master bin file
> vim $(which task-master)
> 
> # change shebang line to include the `-S` flag:
> #!/usr/bin/env -S node --trace-deprecation
> ``
> ```

thanks @simontong!

murphman300:
stellar - thanks folks

tried the native cursor approach to this and already seeing much better results. cant wait to try this tool and see how it improves it further!

Crunchyman-ralph:
This issue has been resolved and will be part of the next release

PR #191, opened 2025-04-12
Fix shebang line for node options

simontong:
Use `-S` flag with `/usr/bin/env` to properly pass options to node.

Addresses - #184

### DIFF
--- a/bin/task-master.js
+++ b/bin/task-master.js
@@ -1,4 +1,4 @@
-#!/usr/bin/env node --trace-deprecation
+#!/usr/bin/env -S node --trace-deprecation
 
 /**
  * Task Master
--- a/README.md
+++ b/README.md
@@ -20,7 +20,13 @@ A task management system for AI-driven development with Claude, designed to work
 
 MCP (Model Control Protocol) provides the easiest way to get started with Task Master directly in your editor.
 
-1. **Add the MCP config to your editor** (Cursor recommended, but it works with other text editors):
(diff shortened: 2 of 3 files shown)


bin/task-master.js
```diff
@@ -1,4 +1,4 @@
-#!/usr/bin/env node --trace-deprecation
+#!/usr/bin/env -S node --trace-deprecation
```
Crunchyman-ralph, line 1: Could you explain what this does exactly ?

Crunchyman-ralph (approved):
lgtm, first rebase, and add a changeset by running:
`npm run changeset`

Crunchyman-ralph:
@simontong can you add a changeset by doing:
```bash
npm run changeset
```

That way it can automatically be part of the changelog that is generated on next release

bin/task-master.js
eyaltoledano: I had added --trace-deprecation for dev purposes. I suspect that if we remove it, this issue ceases to exist

eyaltoledano:
@simontong can you please test by simply removing `--trace-deprecation` from the shebang line? It's meant for dev purposes and I mistakenly checked that in. If it fixes it by removing it, let's not bother with this PR and simply remove the deprecation flag.

Crunchyman-ralph:
Closing this as it tries to resolve the same issue that #243 resolves

Thank you for your contribution!

PR #268, opened 2025-04-18
feat: Enhance remove-task command to handle multiple comma-separated task IDs

Crunchyman-ralph:


Crunchyman-ralph:
Need to implement MCP as well, not just CLI. working on that

PR #302, opened 2025-04-22
Update tutorial.md to use correct npx package name

dschonholtz:
Correct npx package name


eyaltoledano:
@Crunchyman-ralph you'll want to update this too

Crunchyman-ralph:
Fixed in #307

PR #303, opened 2025-04-22
Improve parse-prd UI with realtime feedback

joedanz:
Cleaner PR off of upstream/next


scripts/modules/ai-services.js
```diff
@@ -301,9 +303,10 @@ async function handleStreamingRequest(
 	numTasks,
 	maxTokens,
 	systemPrompt,
-	{ reportProgress, mcpLog, session } = {},
+	{ reportProgress, mcpLog, session, useProgressBar = false } = {},
```
Crunchyman-ralph: why is useProgressBar false by default ?

I think you should keep it, remove that option, and just decide if you should log based on outputFormat and isSilentMode

scripts/modules/task-manager.js
Crunchyman-ralph, line 1: in our effort to slowly refactor, can we not add more to this file.

Can we create something like a src/modules/progress and have something like this:

<details>
<summary>AI Generated response</summary>
To refactor the existing `task-manager.js` file into a more modularized and reusable structure, we can apply a "builder pattern" or a similar design approach. Here's how we can approach this:

### Step 1: Analyze Existing Code
Since the diff for `task-manager.js` is unavailable, I cannot directly analyze its contents. However, if the file manages multiple progress-related functionalities (e.g., PRD progress bar, other types of progress bars), we can start by identifying the commonalities (e.g., rendering, updating progress, events) and differences (e.g., specific configurations or behaviors).

### Step 2: Create `src/modules/progressService`
The goal of the `progressService` module is to encapsulate all progress-related functionality into a reusable and extensible system. It can be structured like this:

1. **Abstract Base Class**: Define a base class `AbstractProgress` for common functionality.
2. **Concrete Implementations**: Create specific progress bar implementations by extending the base class.
3. **Builder Pattern**: Use a builder to configure and instantiate progress bars with a fluent API.

### Example Refactoring Plan
Here’s an idea to modularize progress-related functionality into `progressService`:

---

#### 1. Abstract Base Class (`AbstractProgress`)
Encapsulates common functionality for all progress bars.

```javascript
// src/modules/progressService/AbstractProgress.js
export class AbstractProgress {
  constructor({ max = 100, current = 0, label = '' } = {}) {
    this.max = max;
    this.current = current;
    this.label = label;
  }

  updateProgress(value) {
    this.current = Math.min(value, this.max);
    this.render();
  }

  render() {
    throw new Error('render() must be implemented by subclasses');
  }

  reset() {
    this.current = 0;
    this.render();
  }
}
```

---

#### 2. Concrete Implementations
Create specific progress bar implementations.

**PRD Progress Bar Example:**
```javascript
// src/modules/progressService/PRDProgress.js
import { AbstractProgress } from './AbstractProgress.js';

export class PRDProgress extends AbstractProgress {
  render() {
    console.log(`[PRD Progress]: ${this.label} - ${this.current}/${this.max}`);
  }
}
```

**Other Progress Bar Example:**
```javascript
// src/modules/progressService/OtherProgress.js
import { AbstractProgress } from './AbstractProgress.js';

export class OtherProgress extends AbstractProgress {
  render() {
    console.log(`[Other Progress]: ${this.label} - ${this.current}/${this.max}`);
  }
}
```

---

#### 3. Builder Pattern
Create a builder for easy and consistent creation of progress bars.

```javascript
// src/modules/progressService/ProgressBuilder.js
import { PRDProgress } from './PRDProgress.js';
import { OtherProgress } from './OtherProgress.js';

export class ProgressBuilder {
  constructor() {
    this.type = null;
    this.config = {};
  }

  setType(type) {
    this.type = type;
    return this;
  }

  setMax(max) {
    this.config.max = max;
    return this;
  }

  setLabel(label) {
    this.config.label = label;
    return this;
  }

  build() {
    switch (this.type) {
      case 'PRD':
        return new PRDProgress(this.config);
      case 'Other':
        return new OtherProgress(this.config);
      default:
        throw new Error('Invalid progress bar type');
    }
  }
}
```

---

#### 4. Usage Example
Here’s how the new `progressService` can be used:

```javascript
// Example usage in a script
import { ProgressBuilder } from './modules/progressService/ProgressBuilder.js';

const prdProgress = new ProgressBuilder()
  .setType('PRD')
  .setMax(100)
  .setLabel('Loading PRD Data')
  .build();

prdProgress.updateProgress(30); // Updates progress
prdProgress.updateProgress(80); // Updates progress

const otherProgress = new ProgressBuilder()
  .setType('Other')
  .setMax(50)
  .setLabel('Processing Items')
  .build();

otherProgress.updateProgress(25); // Updates progress
```

---

### Benefits of This Approach
1. **Modularity**: Each progress bar has its own class, making it easy to add new types.
2. **Reusability**: The `AbstractProgress` class provides shared functionality, while concrete subclasses handle specific behaviors.
3. **Flexibility**: The builder pattern allows developers to configure progress bars easily with a fluent API.
4. **Scalability**: Adding new progress types is as simple as creating a new subclass and updating the builder.

### Step 3: Refactor and Migrate
1. Move relevant progress-related functionality from `task-manager.js` into `progressService`.
2. Replace existing progress bar code in `task-manager.js` with the new modularized system.
3. Test thoroughly to ensure all progress-related features work as expected.

Let me know if you'd like further assistance with implementing or refining this refactor!
</details>

scripts/modules/task-manager.js
```diff
@@ -50,200 +57,1712 @@ import {
...
+		onToken: (callback) => emitter.on('token', callback),
+		onTask: (callback) => emitter.on('task', callback),
+		onProgress: (callback) => emitter.on('progress', callback),
+		onComplete: (callback) => emitter.on('complete', callback),
+		onThinking: (callback) => emitter.on('thinking', callback),
+
+		// Cleanup method
+		removeAllListeners: () => emitter.removeAllListeners()
+	};
+}
```
Crunchyman-ralph: this could be included inside the builder pattern I believe

scripts/modules/task-manager.js
```diff
@@ -50,200 +57,1712 @@ import {
...
+ * @param {Object} modelConfig - Model configuration (optional)
+ */
+async function parsePRD(
+	prdPath,
+	tasksPath,
+	numTasks,
+	options = {},
+	aiClient = null,
+	modelConfig = null
+) {
```
Crunchyman-ralph: parsePRD is a new function ?

Instead of adding it here making the file even larger, please move it away, try to prioritize the builder pattern anywhere you see you can, for example the parsePRD can be part of a bigger builder pattern. either way, the refactoring doesn't have to be perfect, but we need to make sure we're not adding to this already large file but instead prioritizing creating one file per "command", which in this case is parsePRD

scripts/modules/ui.js
```diff
@@ -1693,6 +2003,798 @@ async function confirmTaskOverwrite(tasksPath) {
...
+				filledCount++;
+			} else {
+				// This position should be empty
+				progressBar += chalk.gray('░');
+				emptyCount++;
+			}
+		}
+
+		// Use spinner from ora
+		const spinnerFrames = ['⠋', '⠙', '⠹', '⠸', '⠼', '⠴', '⠦', '⠧', '⠇', '⠏'];
```
Crunchyman-ralph: doesn't ora already handle spinning ? why are we doing a spinner ourselves ?

Crunchyman-ralph (changes requested):
Quickly skimmed through the code, I think it needs to be re-read by you checking and cleaning it up.

Based on what you're trying to achieve, there is no way this PR should be 3k lines long.

Try to leverage our existing packages to do what you're triyng to do.

Try to re-use functions as much as possible to create clean code that is readable and that we are able to maintain correctly.

Last thing, see if you can not add to our current huge files and any code you're adding making it a nice addition and making the code better where you touch it.

Like I think most of your additions can allow you to refactor into a different file and import in here making the files potentially smaller.

Will give it a more thourough read, but I really want us to be able to achieve this same functionality with 3x less code, 2.9k is surprising for this feature

PR #345, opened 2025-04-27
Fix MCP expand all tasks file generation

### LINKED ISSUE #344: bug: Task files to not regenerate when expanding all tasks via MCP
### Description

Individual task files do not get regenerated when expanding all tasks via MCP.  Only tasks.json is updated.

### Steps to Reproduce

1. "expand all of my tasks into subtasks based upon the complexity report"

### Expected Behavior

tasks.json should update as well as the individual task files

### Actual Behavior

only tasks.json is being updated

### Environment

- Task Master version: 0.12.1
- Node.js version: 22.15.0
- Operating system: Mac
- IDE (if applicable): Windsurf

joedanz:
Fixes #344

### DIFF
--- a/.changeset/kind-showers-glow.md
+++ b/.changeset/kind-showers-glow.md
@@ -1,0 +1,6 @@
+---
+'task-master-ai': patch
+---
+
+- Fixed file generation after expanding all tasks via MCP
+- Closes #344
--- a/mcp-server/src/core/direct-functions/expand-all-tasks.js
+++ b/mcp-server/src/core/direct-functions/expand-all-tasks.js
@@ -3,165 +3,112 @@
  */
 
 import { expandAllTasks } from '../../../../scripts/modules/task-manager.js';
+import { generateTaskFilesDirect } from './generate-task-files.js';
+import path from 'path';
 import {
 	enableSilentMode,
-	disableSilentMode,
-	isSilentMode
+	disableSilentMode
 } from '../../../../scripts/modules/utils.js';
-import { getAnthropicClientForMCP } from '../utils/ai-client-utils.js';
-import path from 'path';
-import fs from 'fs';
+import { createLogWrapper } from '../../tools/utils.js';
 
 /**
- * Expand all pending tasks with subtasks
+ * Expand all pending tasks with subtasks (Direct Function Wrapper)
  * @param {Object} args - Function arguments
  * @param {string} args.tasksJsonPath - Explicit path to the tasks.json file.
  * @param {number|string} [args.num] - Number of subtasks to generate
- * @param {boolean} [args.research] - Enable Perplexity AI for research-backed subtask generation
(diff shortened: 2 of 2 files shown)


.changeset/kind-showers-glow.md
```diff
@@ -0,0 +1,6 @@
+---
+'task-master-ai': patch
+---
+
+- Fixed file generation after expanding all tasks via MCP
```
Crunchyman-ralph: Remove the dash here, keep it on the second

mcp-server/src/core/direct-functions/expand-all-tasks.js
```diff
@@ -105,6 +105,30 @@ export async function expandAllTasksDirect(args, log, context = {}) {
 				'json' // Use JSON output format to prevent UI elements
 			);
 
+			// After expanding, regenerate individual task files
+			let fileGenResult;
+			try {
+				const { generateTaskFilesDirect } = await import(
```
Crunchyman-ralph: Why this async import ?

Why not import like we import other functions ?

mcp-server/src/core/direct-functions/expand-all-tasks.js
```diff
@@ -116,7 +140,8 @@ export async function expandAllTasksDirect(args, log, context = {}) {
 						prompt: additionalContext,
 						force: forceFlag,
 						tasksExpanded: result.expandedCount,
-						totalEligibleTasks: result.tasksToExpand
+						totalEligibleTasks: result.tasksToExpand,
+						fileGeneration: fileGenResult
```
Crunchyman-ralph: Wondering if the response won't be too long here creating an issue with context on MCP!

Crunchyman-ralph (changes requested):
Maybe remove fileGenResult from returned result, might be too much for context to handle.

Reviewing from the Lebanese moutains on my phone image

mcp-server/src/core/direct-functions/expand-all-tasks.js
joedanz: This is the updated MCP response:
{
  "data": {
    "message": "Successfully expanded all pending tasks with subtasks",
    "details": {
      "research": false,
      "prompt": "",
      "force": true,
      "tasksExpanded": 10,
      "totalEligibleTasks": 10,
      "fileGeneration": {
        "success": true,
        "data": {
          "message": "Successfully generated task files",
          "tasksPath": "/Users/danziger/code/claude-task-master/tasks/tasks.json",
          "outputDir": "/Users/danziger/code/claude-task-master/tasks",
          "taskFiles": "Individual task files have been generated in the output directory"
        },
        "fromCache": false
      }
    }
  }
}

joedanz: Comes back fine in cursor, but whatever you think is best.

eyaltoledano:
@joedanz just one file with conflitcts

PR #349, opened 2025-04-28
Remove external claude usage for MCP 

byingyang:
If you're running task master as an MCP, you're already using an LLM. So why not just use whatever LLM agent calling the MCP to make the calls that you needed an Anthropic API key for? This PR does just that! We now utilize the capability of MCP to expose prompts and so that agent will know how to make the calls itself. No need for any API keys! And as a bonus you can use any LLM instead of just claude.

### DIFF
--- a/mcp-server/src/core/direct-functions/add-task.js
+++ b/mcp-server/src/core/direct-functions/add-task.js
@@ -1,6 +1,6 @@
 /**
  * add-task.js
- * Direct function implementation for adding a new task
+ * Direct function implementation for adding a new task, using FastMCP sampling for AI generation.
  */
 
 import { addTask } from '../../../../scripts/modules/task-manager.js';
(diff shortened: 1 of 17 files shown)


eyaltoledano:
Thanks for the cool PR @byingyang! 

That's a LOT of files changed on your first PR and a really ambitious idea

Could you explain the structural approach to this? What has changed across all of the files? The general approach?

The way I was going to introduce this is through sampling which is already supported via fastmcp. I think I see something in index.js for introducing _prompts_ - how does that solve this?

LMK if you can describe the UX that comes of this too, a bit confused by the agent in the loop additions

Thanks again!

byingyang:
Hey there, created this as a concept and I know it is really nasty. will clean it up soon. But, I am realizing it actually doesn't work with cursor, because cursor does not yet support prompts from MCP; they only support tools. But what is funny is that it still ends up sortof working because there is enough good instructions in the cursor rules for it to guess what things should look like :) 

So I am gonna go back to my original implementation of agent calls a tool to get the prompt to use, executes prompt with context, then executes tools with the result to save tasks, etc. Will also clean up and document how it works. Hopefully someday cursor and others will properly support the protocol...

byingyang:
@eyaltoledano it looks like sampling in fastmcp is a type of tool. I will see if this works!

Crunchyman-ralph:
@byingyang you're using things that don't exist it seems, at least it doesn't seem like it exists when I look at fastMCP documentation

mcp-server/src/core/direct-functions/add-task.js
```diff
@@ -169,85 +138,94 @@ export async function addTaskDirect(args, log, context = {}) {
...
-						system: systemPrompt
-					},
-					{
-						mcpLog: log
-					}
-				);
+				log.info('Initiating FastMCP LLM sampling via client...');
+				// Pass both system and user prompts if the sampling method supports it
+				// Adjust based on actual FastMCP API
+				const completion = await session.llm.complete(userPrompt, { system: systemPrompt });
```
Crunchyman-ralph, line 151: I don't think I've seen this anywhere in the fastMCP documentation.

Crunchyman-ralph (changes requested):
the basis on which this PR is built seems false.

Did you test this before submitting a PR?

mcp-server/src/core/direct-functions/add-task.js
byingyang: you're right it does not exist. used the right one in the parse-prd. still need to fix the rest of them. But also, cursor does not support MCP sampling yet anyways...

byingyang:
When I realized the sampling feature of MCP was the way to go, I redid the branch from scratch, so no. PR is not at all ready anymore.

PR #359, opened 2025-04-30
feat: set httpAgent for Anthropic when socks/https proxy var detected…

aztack:
set httpAgent when `new Anthropic` if process.env.SOCKS_PROXY or process.env.HTTPS_PROXY set


eyaltoledano:
We're not calling the API directly anymore and going through AI SDK 
I'll be adding support for base url which you could use in a bit

Crunchyman-ralph:
This is no longer relevant as of #240

PR #397, opened 2025-05-03
feat: Add .nvmrc and align engines to Node 20

GolanAmir2:
feat: Add .nvmrc and align engines to Node 20

This PR introduces two changes to standardize the Node.js version used for development and contribution:

1.  **Adds a `.nvmrc` file** specifying `20`. This helps developers using NVM easily switch to the intended Node.js version.
2.  **Updates the `engines` field in `package.json`** from `>=14.0.0` to `>=20.0.0`.

**Reasoning:**

*   **Aligns with CI:** The project's continuous integration environment (`.github/workflows/ci.yml`) currently runs tests using Node.js version 20. Aligning the development environment requirements (`engines` and `.nvmrc`) with the CI version ensures consistency and helps prevent compatibility issues.
*   **Compatibility & Testing:** Running tests with older versions like Node 14 may not be reliable or fully supported. Standardizing on Node 20 ensures developers use the same environment where tests are validated.
*   **Node.js Lifecycle:** Node.js 14 and 16 are End-of-Life (EOL), and Node 18 is in maintenance mode.


.changeset/clever-steaks-care.md
Crunchyman-ralph, line 1: this is more of a chore, doesn't need a changeset, please remove

.nvmrc
Crunchyman-ralph, line 1: doesn't it need to be v20 ? maybe we can do something like v22 ?

package.json
```diff
@@ -67,7 +67,7 @@
 		"uuid": "^11.1.0"
 	},
 	"engines": {
-		"node": ">=14.0.0"
+		"node": ">=20.0.0"
```
Crunchyman-ralph: 18 ? lets not be too strict ?

Crunchyman-ralph (changes requested):
Don't know if our node version will actually affect people's projects since they are installing it globally

.nvmrc
GolanAmir2: changed to 22
the prefix v is optional, see nvm --help: 

Node Version Manager (v0.40.1)

Note: <version> refers to any version-like string nvm understands. This includes:
  - full or partial version numbers, starting with an optional "v" (0.10, v0.1.2, v1)

Crunchyman-ralph:
Still hesitant about this PR, cuz if users decide to use this package in their local package manager and decide to use node 16, we will be blocking them from using our package because we require node version 22.

But in reality, we don't need node 22, we don't have that requirement...

Thinking I'm closing this PR for that reason

Crunchyman-ralph:
handled by #612

PR #420, opened 2025-05-05
only ignore `/tasks` folder that is in the main folder 

shaal:
* `/tasks` directory is created by claude-task-master.
* This PR prevent other `tasks/` directories in the repo not getting gitignored


Crunchyman-ralph:
You can always change .gitignore to whatever you want, we're going to keep it that way for now.

Thanks for the though!

PR #438, opened 2025-05-07
feat: Implement custom OpenAI-compatible provider with fallback mechanisms

alicoding:
- Added support for a custom provider in the config manager, including API key and base URL validation.
- Updated supported models JSON to include a custom provider entry.
- Enhanced task manager to handle custom provider hints and validate required environment variables.
- Created test scripts for custom provider object generation and empty fields fix.
- Developed a custom provider implementation using Vercel AI SDK, including text generation and streaming capabilities.
- Introduced model capabilities management to identify models with limited function calling support.
- Added comprehensive error handling and fallback mechanisms for JSON generation when function calling is not supported.


ittarwaqas (approved):

Crunchyman-ralph:
@alicoding can't you just use this #521 that is going to be a part of next release to change the baseUrl to your own server since you are an openAI-compatible provider ? then you can just act exactly the way openai would act making you essentially an openai provider ?

Does that make sense ?

eyaltoledano:
> @alicoding can't you just use this #521 that is going to be a part of next release to change the baseUrl to your own server since you are an openAI-compatible provider ? then you can just act exactly the way openai would act making you essentially an openai provider ?
> 
> Does that make sense ?

i concur

alicoding:
I'm definitely fine with that :) this PR I suppose was created prior to #521 so will use it when it is ready.

Crunchyman-ralph:
awesome dude, but seriously thank you for your contribution

PR #443, opened 2025-05-07
Add pnpm, Bun, and Yarn package‑manager support and publish Python shim for pip

### LINKED ISSUE #229: bug: Support for pnpm
> I have to use pnpm for my monorepo, and unfortunately was unable to make it work. 
 
### Description

We use npm. Users of pnpm can't use it.

We need to support pnpm if it's present.

joedanz:
Requires PyPI token on GitHub, only publishes with leading 'v' on tag.

This code from ci.yml runs earlier at line 50:
      - name: Restore node_modules
        uses: actions/cache@v4
        with:
          path: node_modules
          key: ${{ runner.os }}-node-modules-${{ hashFiles('**/package-lock.json') }}

Resolves #229


python-wrapper/README.md
Crunchyman-ralph, line 1: I honestly don't think we should be exploring the python side of things, lets stick to our current javascript realm for now.

Didn't see anyone requesting it, would rather not include this. Please remove from this PR

joedanz:
> Thats a good question, do we want to support python?

It’s really just a wrapper and gives us the ability to reach out to the python community much more easily.  To me its more why not, since the barrier to entry is so low without even touching the actual code.

bun.lock
Crunchyman-ralph, line 1: having a bun.lock simply means our project's package manager is bun.

Doesn't mean we will be able to support bun.

Our project needs to have one lock file or else it will conflict with other package managers.

So now when I try to run npm install, npm will complain that there are a bunch of lock files, and they will inevitably be stale since we're only doing `npm i`. With your implementation, it will mean that we will now have to do a pnpm install, npm install, bun install, etc...

I think supporting multiple registries (yarn, npm, pnpm, bun) is just a matter of putting our package in their respective registries.

And maybe you can do that with changesets and the CI, as simple as that.

Adding these lockfiles won't allow us to be compatible with them, probably the contrary

pnpm-lock.yaml
Crunchyman-ralph, line 1: same as bun.lock comment

README.md
```diff
@@ -64,13 +64,13 @@ Can you help me expand task 4?
...
-# OR install locally within your project
-npm install task-master-ai
-```
+| Package manager | Global install                   | Project install           | One‑shot (npx‑style)         |
+| --------------- | -------------------------------- | ------------------------- | ---------------------------- |
+| **npm**         | `npm i -g task-master-ai`        | `npm i task-master-ai`    | `npx -y task-master-ai`      |
+| **pnpm**        | `pnpm add -g task-master-ai`     | `pnpm add task-master-ai` | `pnpm dlx -y task-master-ai` |
+| **Bun**         | `bun add -g task-master-ai`      | `bun add task-master-ai`  | `bunx -y task-master-ai`     |
+| **Yarn**        | `yarn global add task-master-ai` | `yarn add task-master-ai` | `yarn dlx task-master-ai`    |
+| **pip / pipx**  | `pip install task-master-ai`     | _(not applicable)_        | `pipx run task-master-ai`    |
```
Crunchyman-ralph, line 73: lets keep this, but first make it compatible with those packages

.github/workflows/ci.yml
```diff
@@ -93,3 +109,30 @@ jobs:
...
+        run: python -m pip install build
+
+      - name: Build Python wheel and sdist
+        working-directory: ./python-wrapper
+        run: python -m build
+
+      - name: Publish package to PyPI
+        uses: pypa/gh-action-pypi-publish@release/v1
+        with:
+          password: ${{ secrets.PYPI_TOKEN }}
```
Crunchyman-ralph: lets not do that

.github/workflows/ci.yml
```diff
@@ -61,18 +61,34 @@ jobs:
...
+          if [ "${{ matrix.pm }}" = "npm" ]; then
+            npm ci
+          elif [ "${{ matrix.pm }}" = "pnpm" ]; then
+            pnpm setup
+            pnpm install --frozen-lockfile
+          elif [ "${{ matrix.pm }}" = "bun" ]; then
+            bun install --frozen-lockfile
+          elif [ "${{ matrix.pm }}" = "yarn" ]; then
+            yarn install --frozen-lockfile
+          fi
```
Crunchyman-ralph: lets not do that

.github/workflows/ci.yml
```diff
@@ -61,18 +61,34 @@ jobs:
   test:
     needs: setup
     runs-on: ubuntu-latest
+    strategy:
+      matrix:
+        pm: [npm, pnpm, bun, yarn]
```
Crunchyman-ralph: lets explore the CI implications once we figure out the core of what this PR is supposed to do.

Crunchyman-ralph (changes requested):
lets re-explore what is actually required to be compatible with bun, pnpm, yarn.

I don't think this is it.

.github/workflows/ci.yml
joedanz: ok reverted back to original ci.yml

.github/workflows/ci.yml
joedanz: reverted back to original ci.yml

.github/workflows/ci.yml
joedanz: reverted back to orig

bun.lock
joedanz: ok removed the other lock files, thx.

Crunchyman-ralph (changes requested):
Lets remove python, we will re-visit at a later time

PR #453, opened 2025-05-09
Enhance AI services integration by adding Cursor provider support and updating glossary.

tboy1337:
- Added Cursor AI provider to `ai-services-unified.js` with functions for text generation and streaming.
- Updated error handling in `isRetryableError` and `_extractErrorMessage` functions for improved robustness.
- Expanded `supported-models.json` to include Cursor AI model details.
- Revised glossary in `glossary.mdc` to reference new Cursor AI integration.


eyaltoledano:
Did you test and verify this is working @tboy1337?

I’m sceptical about the implementation of cursor as a provider

tboy1337:
I can't currently test it as i am travelling, if someone else could test it that would be great.

Crunchyman-ralph:
@tboy1337 I don't really understand what you're saying.

You created a PR that you didn't test ? How do you post this PR if you don't even know if it works ?

tboy1337:
Even if it isn't working it will be 90% there so it's worth posting imo as this feature is badly needed, probably should have been a draft PR but there you go.  Instead of taking the effort to post maybe put that little extra effort in and test it xD

Crunchyman-ralph:
Great!

Let me know once you've tested it and I'll be sure to review and test on my side!

tboy1337:
I just tested it and its working xD

tboy1337:
I think this is going to break it for everyone thats not using cursor though.

Crunchyman-ralph:
Try to undo your last commit as it tries to change the package.json scripts and default task master config which isn't something we want to change

eyaltoledano:
If this works (I’m dumbfounded this would be posted without checking it actually works) it will need to be properly added to the setup for models with conditional logiv not to display cursor as a provider if terminal’s not in the cursor agent. 

Don’t love the idea of passing it as a provider tbh

Crunchyman-ralph:
I saw another PR use something like `shx` so that it works with windows and unix

tboy1337:
You can only use bash on Windows if WSL is installed and that isnt by default on Windows installs, the current method works regardless.

Karnith:
I tested this pr and am unable to get it to work. I'm using Windows 11 and git bash. It seems that the globa1This in your code doesn't surface the __Cursor__ object attribute when printed out...
My cursor info:
Version: 0.50.4 (user setup)
VSCode Version: 1.96.2
Commit: 8ea935e79a50a02da912a034bbeda84a6d3d3550
Date: 2025-05-14T16:17:45.517Z
Electron: 34.3.4
Chromium: 132.0.6834.210
Node.js: 20.18.3
V8: 13.2.152.41-electron.0
OS: Windows_NT x64 10.0.26100

I feel that this would be a huge advantage to be able to use the cursor AI agent, which in turn uses the model defined instead of setting the same model in env, especially when payment is made to cursor for the model use. There is also the slow request feature in cursor that allows you to use a model when the 500 requests are used, saving money as well.

Crunchyman-ralph:
Thanks for your contribution @tboy1337, closing this for now, we're so close though, going to be a value PR to come back to!

PR #478, opened 2025-05-11
Fix CLI --force flag on parse-prd command

### LINKED ISSUE #477: bug: --force flag not being respected on CLI parse-prd command
### Description

The --force flag is not respected when running `parse-prd`, causing the command to prompt for confirmation or fail even when --force is provided.

### Steps to Reproduce

taskmaster parse-prd scripts/prd.txt --force

### Expected Behavior

Immediately parse PRD with no confirmation

### Actual Behavior

Confirmation appears.

joedanz:
Fix CLI --force flag for parse-prd command

Previously, the --force flag was not respected when running `parse-prd`, causing the command to prompt for confirmation or fail even when --force was provided. This patch ensures that the flag is correctly passed and handled, allowing users to overwrite existing tasks.json files as intended.

- Fixes #477


Crunchyman-ralph (approved):
tested, works. LGTM

PR #574, opened 2025-05-22
Fix parse-prd --append when using the command line and an input PRD

sleeper:
`parsePRD` is expecting options to be called `append` and `force` and not `useAppend` and `useForce`.
Seems bug has been introduced with commit 0288311
.
In practice it means that without this a command like `task-master parsePRD --input script/my-feeature-PRD.md --append` is failing with a message like:

`[ERROR] Output file tasks/tasks.json already exists. Use --force to overwrite or --append.`.


sleeper:
Oops !  Sorry for the double work!

Issue #561, opened 2025-05-21
bug: Support AWS bedrock - not addressed with v0.14 release


mspiegel31:
### Description

[this comment](https://github.com/eyaltoledano/claude-task-master/issues/287#issuecomment-2842475846) states that #287 should be closed with the release of version `0.14.x`. this does not appear to be the case.

There are a few key differences between how the Anthropic API functions and the AWS Bedrock API functions.  Notably:

1. AWS bedrock requires the usage of a either a [credentials chain](https://ai-sdk.dev/providers/ai-sdk-providers/amazon-bedrock#using-aws-sdk-credentials-chain-instance-profiles-instance-roles-ecs-roles-eks-service-accounts-etc) (preferred) or [static tokens](https://ai-sdk.dev/providers/ai-sdk-providers/amazon-bedrock#provider-instance).  These links from the vercel `ai-sdk` delve into specific examples for consuming these authn techniques
2. AWS bedrock models **do not** follow the same naming conventions as the anthropic API (see [here](https://docs.anthropic.com/en/api/claude-on-amazon-bedrock#api-model-names) for naming conventions).  Note that AWS bedrock models use a version suffix (i.e. `-v1:0`) in their model references

### Steps to Reproduce

1. run `task-master init`
1. modify `.taskmasterconfig` with sample configuration:
```json
{
  "models": {
    "main": {
      "provider": "anthropic",
      "modelId": "claude-3-7-sonnet-20250219-v1:0",
      "maxTokens": 120000,
      "temperature": 0.2,
      "baseUrl": "https://bedrock.us-east-1.amazonaws.com"
    },
    "research": {
      "provider": "openrouter",
      "modelId": "perplexity/sonar-pro",
      "maxTokens": 8700,
      "temperature": 0.1,
      "baseUrl": "https://openrouter.ai/api/v1"
    },
    "fallback": {
      "provider": "anthropic",
      "modelId": "claude-3-5-sonnet-20241022-v2:0",
      "maxTokens": 8192,
      "temperature": 0.1,
      "baseUrl": "https://bedrock.us-east-1.amazonaws.com"
    }
  },
  "global": {
    "logLevel": "info",
    "debug": false,
    "defaultSubtasks": 5,
    "defaultPriority": "medium",
    "projectName": "Taskmaster",
    "ollamaBaseUrl": "http://localhost:11434/api",
    "azureOpenaiBaseUrl": "https://your-endpoint.openai.azure.com/",
[...]

PR #585, opened 2025-05-23
Add Amazon Bedrock support

dgiberso:
Adding Amazon Bedrock support. See https://github.com/eyaltoledano/claude-task-master/issues/561


Crunchyman-ralph:
@dgiberso  I have added bedrock support here #607 feel free to let me know if this is enough for you

dgiberso:
Much better solution. Thank you.

PR #593, opened 2025-05-25
fix(add-task): removes stdout in add-task which will crash MCP server

eyaltoledano:
fixes an issue in which stdout leaks out of add-task causing the mcp server to crash if used.


Crunchyman-ralph (changes requested):
Please add a changeset, and lets release!

This is a patch btw

PR #600, opened 2025-05-26
Add Azure OpenAI support.

Bovey0809:
Add provider for Azure openai.


.cursor/mcp.json
eyaltoledano, line 1: why is this deleted? don’t check this change in this PR plz

scripts/modules/ai-services-unified.js
eyaltoledano, line 1: Why both azure and azureOpenAi?

scripts/modules/ai-services-unified.js
```diff
@@ -189,6 +203,7 @@ function _resolveApiKey(providerName, session, projectRoot = null) {
 		perplexity: 'PERPLEXITY_API_KEY',
 		mistral: 'MISTRAL_API_KEY',
 		azure: 'AZURE_OPENAI_API_KEY',
+		'azure-openai': 'AZURE_OPENAI_API_KEY',
```
eyaltoledano, line 206: Why duplicate keys?

scripts/modules/supported-models.json
eyaltoledano, line 1: Looks like duplicate entries in azure

eyaltoledano:
Added some notes
Not sure why theres both azure and azureOpenAi
Lots of duplication

please also avoid the changes to the way the config works, esp deleting taskmasterconfig and mcp.jaon is a no go

PR #607, opened 2025-05-26
Feat/add.azure.and.other.providers

Crunchyman-ralph:


scripts/modules/supported-models.json
sebastienfi, line 1: @Crunchyman-ralph I think you forgot to push your changes to this file. It doesn't have a `vertex` key, so the feature is not usable.

Bovey0809:
Great, any example for the Azure configs? @Crunchyman-ralph

PR #625, opened 2025-05-31
Release 0.16.0

Crunchyman-ralph:


CONTRIBUTING.md
```diff
@@ -0,0 +1,335 @@
...
+- Add model validation for custom Ollama endpoints
+- Update configuration schema to support custom models
+- Add tests for new validation logic"
+```
+
+## 🔧 Development Setup
+
+### Prerequisites
+
+- Node.js 14+
```
eyaltoledano: 18?

CONTRIBUTING.md
```diff
@@ -0,0 +1,335 @@
+# Contributing to Task Master
```
eyaltoledano, line 1: In future versions we could also ship this as taskmaster_contributions.mdc which can be included optionally through init if the user says they want to contirbute to taskmaster itself. and then the AI agent would be able to identify improvements, code them up and submit PRs through the user.

In effect, a method for decentralizing not only ideas and bugs from humans using cursor, but also complete PRs addressing issues users may run into (in real-time) submitted by the AI agents themselves (provided the primitives to do so easily)

could even be a script. will add as an idea/issue

CONTRIBUTING.md
```diff
@@ -0,0 +1,335 @@
...
+├── src/                   # Shared utilities and providers and well refactored code (we are slowly moving everything here)
+├── tests/                 # Test files
+├── docs/                  # Documentation
+└── .cursor/               # Cursor IDE rules and configuration
+└── assets/							   # Assets like rules and configuration for all IDEs
+```
+
+### Key Areas for Contribution
+
+- **CLI Commands**: `scripts/modules/commands.js`
```
eyaltoledano, line 293: We should add "out of scope for now" 

and include UI in there for example

README.md
```diff
@@ -109,15 +109,18 @@ Change the main, research and fallback models to <model_name>, <model_name> and
...
 ```
 
-#### 5. Make sure you have a PRD in `<project_folder>/scripts/prd.txt`
+#### 5. Make sure you have a PRD
 
-An example of a PRD is located into `<project_folder>/scripts/example_prd.txt`.
+For **new projects**: Create your PRD at `.taskmaster/docs/prd.txt`  
+For **existing projects**: You can use `scripts/prd.txt` or migrate with `task-master migrate`
+
+An example PRD template is available after initialization in `.taskmaster/templates/example_prd.txt`.
```
eyaltoledano, line 123: We can now start projects by adding a task without a PRD -- this should probably be mentioned

docs/migration-guide.md
```diff
@@ -0,0 +1,235 @@
+# Migration Guide: New .taskmaster Directory Structure
+
+## Overview
+
+Task Master v3.x introduces a new `.taskmaster/` directory structure to keep your project directories clean and organized. This guide explains the benefits of the new structure and how to migrate existing projects.
```
eyaltoledano: what is v3.x?

docs/migration-guide.md
```diff
@@ -0,0 +1,235 @@
...
+
+## What's New
+
+### Before (Legacy Structure)
+
+```
+your-project/
+├── tasks/                    # Task files
+│   ├── tasks.json
+│   ├── task-1.md
```
eyaltoledano: task files were in .txt format not md

docs/migration-guide.md
```diff
@@ -0,0 +1,235 @@
...
+
+### After (New Structure)
+
+```
+your-project/
+├── .taskmaster/              # Consolidated Task Master files
+│   ├── config.json          # Configuration (was .taskmasterconfig)
+│   ├── tasks/               # Task files
+│   │   ├── tasks.json
+│   │   ├── task-1.md
```
eyaltoledano: still not md, will stay txt

docs/migration-guide.md
eyaltoledano, line 1: I understand we had to move fast, but I feel like the entire migration could have been abstracted away from the user experience altogether and operated as a background process if legacy markers are detected in the project

That would avoid the need to create a whole new command that doesn't actually have any production functionality for the user (i.e. migrating/upgrading software isn't software functionality the user cares about)

Don't wanna hold this release up, but at the same time, this could be done automatically by Taskmaster. I think us having to have a migration guide is a good example of introducing complexity to the user experience that we wanna avoid in favor of a "just works" approach

As an example, the script as it is right is is perfect
But instead of relying on the user to either trigger the command and/or manually make the migration

We could have middleware logic at the CLI and MCP layer triggered anytime a command or tool is called that detects legacy structure and automatically tacks on a second stage for that command whch asks the user if they'd like to upgrade to the new taskmaster config, and that the agent can perform the migration for them (much in the way we have it now)

Maybe this is just a question of detection and documentation 
-> with detection, the CLI or the MCP could simply ask the user if they want to migrate so the user never really needs to think about the migration or if there is a migration to do or something. the agent can detect its needed, ask for a confirmation, and then just call the command and do it for the user

wanna minimize exposure to complexity as much as possible

might be something for us to improve next release

mcp-server/src/core/direct-functions/initialize-project.js
```diff
@@ -75,7 +75,7 @@ export async function initializeProjectDirect(args, log, context = {}) {
...
+				'Now that the project is initialized, the next step is to create the tasks by parsing a PRD. This will create the tasks folder and the initial task files (tasks folder will be created when parse-prd is run). The parse-prd tool will require a prd.txt file as input (typically found in the project root directory, .taskmaster/docs directory). You can create a prd.txt file by asking the user about their idea, and then using the scripts/example_prd.txt file as a template to genrate a prd.txt file in scripts/. You may skip all of this if the user already has a prd.txt file. You can THEN use the parse-prd tool to create the tasks. So: step 1 after initialization is to create a prd.txt file in .taskmaster/docs/prd.txt or confirm the user already has one. Step 2 is to use the parse-prd tool to create the tasks. Do not bother looking for tasks after initialization, just use the parse-prd tool to create the tasks after creating a prd.txt from which to parse the tasks. You do NOT need to reinitialize the project to parse-prd.',
```
eyaltoledano: adjustment needed

isn't this supposed to be
scripts/example_prd.txt -> .taskmaster/templates/example_prd.txt?

`to genrate a prd.txt file in scripts/` to `to generate a prd.txt file in .taskmaster/docs/prd.txt`

mcp-server/src/core/direct-functions/parse-prd.js
eyaltoledano, line 1: where does the new project root override come into play? resolveProjectPath?

mcp-server/src/tools/analyze.js
```diff
@@ -94,11 +95,7 @@ export function registerAnalyzeProjectComplexityTool(server) {
 
 				const outputPath = args.output
 					? path.resolve(args.projectRoot, args.output)
-					: path.resolve(
-							args.projectRoot,
-							'scripts',
-							'task-complexity-report.json'
-						);
+					: path.resolve(args.projectRoot, COMPLEXITY_REPORT_FILE);
```
eyaltoledano, line 98: when you're calling projectRoot here, is this going to be properly using the override you just introduced? idk if you're using the util so it wouldn't go through the 5 step process no?

mcp-server/src/tools/utils.js
```diff
@@ -588,31 +589,52 @@ function withNormalizedProjectRoot(executeFn) {
...
 		try {
-			// Determine raw root: prioritize args, then session
-			let rawRoot = args.projectRoot;
-			if (!rawRoot) {
-				rawRoot = getRawProjectRootFromSession(session, log);
-				rootSource = 'session';
-			} else {
-				rootSource = 'args';
+			// PRECEDENCE ORDER:
+			// 1. TASK_MASTER_PROJECT_ROOT environment variable (from process.env or session)
```
eyaltoledano, line 593: roger roger, i see it now

scripts/init.js
```diff
@@ -286,10 +297,7 @@ function copyTemplateFile(templateName, targetPath, replacements = {}) {
 
 			if (newLines.length > 0) {
 				// Add a comment to separate the original content from our additions
-				const updatedContent =
-					existingContent.trim() +
-					'\n\n# Added by Claude Task Master\n' +
-					newLines.join('\n');
+				const updatedContent = `${existingContent.trim()}\n\n# Added by Claude Task Master\n${newLines.join('\n')}`;
```
eyaltoledano: let's remove "Claude" from the name lol

by Taskmaster AI

scripts/init.js
```diff
@@ -500,44 +505,44 @@ function createProjectStructure(addAliases, dryRun) {
...
 		'dev_workflow.mdc',
-		path.join(targetDir, '.cursor', 'rules', 'dev_workflow.mdc')
+		path.join(targetDir, '.cursor/rules/dev_workflow.mdc')
 	);
 
 	// Copy taskmaster.mdc
 	copyTemplateFile(
 		'taskmaster.mdc',
-		path.join(targetDir, '.cursor', 'rules', 'taskmaster.mdc')
+		path.join(targetDir, '.cursor/rules/taskmaster.mdc')
```
eyaltoledano, line 533: we could totally adjust these so all of our rules go into .cursor/rules/taskmaster/rule_name.mdc so our rules are tucked into a single folder. this would be a good opportunity to throw this in, very easy improvement

scripts/init.js
```diff
@@ -638,18 +632,22 @@ function createProjectStructure(addAliases, dryRun) {
 		);
 	} else if (dryRun) {
 		log('info', 'DRY RUN: Skipping interactive model setup.');
+	} else if (options?.yes) {
+		log('info', 'Skipping interactive model setup due to --yes flag.');
+		log(
+			'info',
+			'You can configure AI models later using "task-master models --setup" or "task-master models --set-..." commands.'
```
eyaltoledano: this should mention that the default models/config will be used until the setup is used

scripts/init.js
```diff
@@ -664,76 +662,27 @@ function createProjectStructure(addAliases, dryRun) {
...
-					) +
-					'\n' +
-					chalk.dim(
-						'* Use the task-master command without arguments to see all available commands.'
-					),
+				`${chalk.cyan.bold('Things you should do next:')}\n\n${chalk.white('1. ')}${chalk.yellow(
+					'Configure AI models (if needed) and add API keys to `.env`'
+				)}\n${chalk.white('   ├─ ')}${chalk.dim('Models: Use `task-master models` commands')}\n${chalk.white('   └─ ')}${chalk.dim(
+					'Keys: Add provider API keys to .env (or inside the MCP config file i.e. .cursor/mcp.json)'
+				)}\n${chalk.white('2. ')}${chalk.yellow(
```
eyaltoledano, line 669: we probably should mention you can just add-task

scripts/modules/ai-services-unified.js
```diff
@@ -19,18 +19,41 @@ import {
...
-import * as anthropic from '../../src/ai-providers/anthropic.js';
-import * as perplexity from '../../src/ai-providers/perplexity.js';
-import * as google from '../../src/ai-providers/google.js';
-import * as openai from '../../src/ai-providers/openai.js';
-import * as xai from '../../src/ai-providers/xai.js';
-import * as openrouter from '../../src/ai-providers/openrouter.js';
-import * as ollama from '../../src/ai-providers/ollama.js';
-// TODO: Import other provider modules when implemented (ollama, etc.)
+// Import provider classes
+import {
```
eyaltoledano, line 31: this is 🔥

scripts/modules/config-manager.js
```diff
@@ -125,6 +127,15 @@ function _loadAndValidateConfig(explicitRoot = null) {
 			};
 			configSource = `file (${configPath})`; // Update source info
 
+			// Issue deprecation warning if using legacy config file
+			if (isLegacy) {
+				console.warn(
+					chalk.yellow(
+						`⚠️  DEPRECATION WARNING: Found configuration in legacy location '${configPath}'. Please migrate to .taskmaster/config.json. Run 'task-master migrate' to automatically migrate your project.`
```
eyaltoledano, line 134: bravo

scripts/modules/task-manager/migrate.js
eyaltoledano, line 1: elegant

src/ai-providers/google.js
eyaltoledano, line 1: we removed this -- i see google-vertex
but aren't they two separate providers? gemini api connection direct vs. vertex setup? is the regular gemini provider still working?

src/constants/paths.js
eyaltoledano, line 1: is this exposed this way for convenience or to let users adjust if they wanna put stuff in other places?

eyaltoledano (changes requested):
> [!IMPORTANT]
> BLOCKERS
> - `taskmaster.mdc` and `dev_workflow.mdc` need to be updated based on the new architecture else the agent might use incorrect arguments which would degrade UX

- my main piece of feedback is that with large changes to structure, we need to also adjust our own Cursor rules in .cursor/rules -- right now none of them have this update. i suggest using the branch as an opportunity to feed its changes to the AI and requesting any updates/edits across our rule files, especially dev_workflow.mdc and taskmaster.mdc as they are consumer-facing and need to be adjusted or the agent will be confused as to why its arguments arent working.
- some docs still reference old ways of doing things
- parse prd system prompt update references old config path setup
- unclear whether the new project root override is properly integrated. seeing lots of `args.projectRoot` on their own, which means it's only using the projectRoot from the arg and not the override if it's mentioned or if the args.projectRoot is not present. Just gotta make sure that this override is properly used everywhere. tough to do considering how many places root is mentioned
- i noticed we removed the google.js provider in favor of google-vertex -- but they are two separate providers AFAIK. I left a comment please clarify
- migrate script is 👌 
- really nicely set up the provider class, well done
- really appreciate the unit tests now including the commands separately

Overall, 95%+ there minus my nitpicks above

But definitely need to adjust the cursor rules

1. internal cursor rules (taskmaster's) for how the new architecture works. literally most rules will need adjusting for sure
2. the end-consumer rules we ship with init `taskmaster.mdc` and `dev_workflow.mdc` absolutely need updates since the ai agent needs to be aware of the new setup.

docs/migration-guide.md
eyaltoledano: now that i've completed my review, I confirm this is mostly a .mdc rule update that would fix this since the legacy markers DO get found -> we jsut need to let the agent know so the agent can be like "hey uh btw do you wanna migrate? i can do this for you rn in 2 sec if you want"

vs. the user having to deal with a migration altogether

src/ai-providers/google.js
rtmcrc: yes Gemini is highly needed

mcp-server/src/core/direct-functions/initialize-project.js
Crunchyman-ralph: good catch

CONTRIBUTING.md
Crunchyman-ralph: yup

CONTRIBUTING.md
Crunchyman-ralph: We'll add it when we get there, we don't want to give our contributors the wrong idea

README.md
Crunchyman-ralph: done

scripts/modules/ai-services-unified.js
Crunchyman-ralph: Thank you, thank you

scripts/modules/config-manager.js
Crunchyman-ralph: backwards compatibility didn't want breaking changes

src/ai-providers/google.js
Crunchyman-ralph: I don't see it removed anywhere, still being used in ai-services-unified. We need google for the normal gemini stuff, and vertex is the "bedrock" of google's hosting services. So they are 2 separate things.

mcp-server/src/tools/analyze.js
Crunchyman-ralph: I see you saw the wrapped withNormalizedProjectRoot that covers this use case so we can safely use args.projectRoot without the headaches

mcp-server/src/tools/utils.js
Crunchyman-ralph: yup

scripts/init.js
Crunchyman-ralph: hahaha

scripts/init.js
Crunchyman-ralph: don't want to mess with joe's PR, and I also think there is no harm in duplication, especially since we have our own task-master rules and our client-facing task-master rules, its a coincidence that we also use it because our product's nature is like that, but I like the clear separation of concerns. The duplication is appreciated in this case.

scripts/init.js
Crunchyman-ralph: done

src/constants/paths.js
Crunchyman-ralph: This stuff is defined in one place so that if one day, for some reason, we want to change the locations and not have it be .taskmaster but instead something else in the future, we can.

So that it would avoid the headache that I had to have to findAll of all instances of scripts/prd.txt and have to change them one by one.

At least here is I decide prds don't live in /docs/prd.txt but in /prds/prd.txt, we can do that change very easily

scripts/init.js
Crunchyman-ralph: done

src/ai-providers/google.js
eyaltoledano: confirmed, all good.

eyaltoledano (approved):
approved pending requested changes

docs/migration-guide.md
Crunchyman-ralph: oops

docs/migration-guide.md
Crunchyman-ralph: done

PR #626, opened 2025-05-31
feat: added home brew support

nickiiitu:


nickiiitu:
hi @Crunchyman-ralph i wnated to contribute towards your project , i have added homebrew support for task master , can you review this and help me if somethings needs to be added

Formula/task-master.rb
```diff
@@ -0,0 +1,21 @@
...
+  
+    def install
+      system "npm", "install", *std_npm_args
+      bin.install_symlink libexec/"bin/task-master" => "task-master"
+      bin.install_symlink libexec/"bin/task-master-mcp"
+    end
+  
+    test do
+      system bin/"task-master", "--version"
+      assert_match "0.15.0", shell_output("#{bin}/task-master --version")
```
xkcoding, line 18: I noticed that in the `Formula/task-master.rb` file, the version is currently hardcoded as “0.15.0.” I wonder if it might be more elegant to use GitHub Actions to dynamically update the formula with the latest version. This approach could help ensure consistency and reduce the need for manual updates.

Thank you for your hard work on this feature—it’s a great addition! I hope my suggestion is helpful. 😊

Crunchyman-ralph:
@xkcoding wondering if we can do some sort of automatic release with changeset.

Otherwise, this feels like too much of a hassle for not much value

xkcoding:
@Crunchyman-ralph I think it's possible to use GitHub Actions to listen for tags, similar to: https://github.com/marketplace/actions/github-tag

Crunchyman-ralph:
oh nice, @nickiiitu if we're going to do this, might as well do it well, we already have github actions and use changeset for npm releases, think we can include a github action that handles releasing on homebrew ?

NikhilTheFlyter:
sure will integrate github actions by this weekend

nickiiitu:
sure will do it

Crunchyman-ralph:
Getting stale and honestly don't want to also support homebrew just yet.

Will add homebrew once we have global config figured out, makes more sense that way!

PR #628, opened 2025-05-31
Add Requesty models provider

rtmcrc:
This PR adds support for Requesty AI models provider.


.changeset/itchy-eagles-roll.md
```diff
@@ -0,0 +1,10 @@
+---
+"task-master-ai": minor
```
Crunchyman-ralph, line 2: this is a major
```suggestion
"task-master-ai": major
```

Since its a feature

Crunchyman-ralph (changes requested):
Small nitpick, else everything else looks good!

PR #641, opened 2025-06-02
fix: bedrock set model and other fixes

Crunchyman-ralph:
- Closes #641


Crunchyman-ralph (commented):
self-reviewed, auto-merging

PR #659, opened 2025-06-03
Add claude GitHub actions 1748965137427

HarleyCoops:


package.json
```diff
@@ -67,6 +67,7 @@
 		"ollama-ai-provider": "^1.2.0",
 		"openai": "^4.89.0",
 		"ora": "^8.2.0",
+		"task-master-ai": "^0.15.0",
```
Crunchyman-ralph, line 70: remove this

.cursor/mcp.json
Crunchyman-ralph, line 1: remove diff here, you're not actually changing anything

.github/workflows/claude.yml
Crunchyman-ralph, line 1: what is this workflow trying to do ?

.gitignore
```diff
@@ -64,3 +64,17 @@ dev-debug.log
...
+.vscode
+*.suo
+*.ntvs*
+*.njsproj
+*.sln
+*.sw?
+# OS specific
+# Task files
+tasks.json
+tasks/
```
Crunchyman-ralph, line 80: I don't think we want to ignore any of these ?

.roo/rules/ai_providers.md
eyaltoledano, line 1: why are we passing these roo rules? they're internal.

PR #723, opened 2025-06-09
Fix Cursor deeplink installation with copy-paste instructions

joedanz:
## Problem
The Cursor one-click install button wasn't working on GitHub because:
- GitHub strips `cursor://` protocol links for security reasons
- Clickable deeplinks don't work in GitHub markdown for security

## Solution
- ✅ **Replaced clickable button with copy-paste instructions** that work reliably on GitHub
- ✅ **Used `text` code block** to ensure GitHub adds a native copy button automatically
- ✅ **Added clear visual instructions** with clipboard emoji for better UX

Fixes the installation flow by providing a reliable, GitHub-compatible solution.


Crunchyman-ralph (approved):
lgtm

PR #820, opened 2025-06-19
feat: add HTTP transport option for running taskmaster

DanielPolatajko:
## Description

This PR adds an HTTP server option for running taskmaster. I was motivated to make this change because I like taskmaster but I often develop in a devcontainer setup where adding a taskmaster container to my docker compose and accessing over local network would be preferable to using stdio transport.

## Type of Change

- [ ] Bug fix
- [x] New feature
- [ ] Breaking change
- [ ] Documentation update

## Testing

- [x] All existing tests pass
- [x] I have added tests for new functionality

## Changeset

## Additional Notes

I guess there are other considerations relating to things like file storage on the container etc which would be relevant to this feature, but I don't think they're blocking. I'll be happy to add some docs about running taskmaster in a containerised setup if you think its useful.

### DIFF
--- a/.changeset/bumpy-hornets-yell.md
+++ b/.changeset/bumpy-hornets-yell.md
@@ -1,0 +1,5 @@
+---
+"task-master-ai": minor
+---
+
+Add an HTTP server with executable for streaming and SSE connections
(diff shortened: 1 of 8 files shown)


Crunchyman-ralph:
Please point to next branch as per contributing.md guidelines :D

Crunchyman-ralph:
not yet the moment for http transport, but soon.

Closing this for now

PR #833, opened 2025-06-20
Call rules interactive setup during init

joedanz:
Call rules interactive setup during taskmaster init.
Does not get called when --rules flag is used to specify rule profiles for init


Crunchyman-ralph:
works really well, merging

PR #835, opened 2025-06-20
fix: Store tasks in Git by default

joedanz:
Re-implement default behavior of storing tasks in Git (reverted in PR 743)


Crunchyman-ralph (commented):
nice

PR #836, opened 2025-06-20
experimental: try to remove --package and just do npx task-master-ai

Crunchyman-ralph:


Crunchyman-ralph:
favoring #1087 instead (tsup)

PR #846, opened 2025-06-22
Default to Cursor profile for MCP init when no rules specified

joedanz:
### Summary
Updates MCP `initialize_project` to default to Cursor profile instead of installing all rule profiles when no profiles specified.

### Changes
- **MCP initialization**: Defaults to `["cursor"]` rules when none specified
- **CLI behavior**: Unchanged - runs interactive rules setup when no profiles specified
- **Impact**: Cleaner project directory without extra rule profiles installed by default

### DIFF
--- a/.changeset/mean-papayas-share.md
+++ b/.changeset/mean-papayas-share.md
@@ -1,0 +1,5 @@
+---
+"task-master-ai": patch
+---
+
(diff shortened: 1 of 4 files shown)


Crunchyman-ralph (approved):
lgtm

PR #852, opened 2025-06-23
fix: Subtask generation fails on gemini-2.5-pro

hrmshandy:
## Description

This PR fixes a critical issue where subtask generation fails on gemini-2.5-pro unless explicitly prompted to `return 'details' field as a string not an object`

## Type of Change

- [x] Bug fix
- [ ] New feature
- [ ] Breaking change
- [ ] Documentation update

## Testing

- [x] All existing tests pass
- [ ] I have added tests for new functionality

## Changeset

### DIFF
--- a/.changeset/tame-vans-throw.md
+++ b/.changeset/tame-vans-throw.md
@@ -1,0 +1,5 @@
+---
+"task-master-ai": patch
+---
+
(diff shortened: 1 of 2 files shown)


Crunchyman-ralph (changes requested):
add a changeset using `npm run changeset` so that we can see your fix in upcoming releases and credit!

Crunchyman-ralph (approved):
lgtm

PR #872, opened 2025-06-25
Feat: Implemented advanced settings for Claude Code AI provider

geoh:
## Description

Added advanced settings for Claude Code AI provider in base config file. Including command specific overrides.

## Type of Change

- [ ] Bug fix
- [x] New feature
- [ ] Breaking change
- [ ] Documentation update

## Testing

- [x] All existing tests pass
- [x] I have added tests for new functionality

## Changeset

## Additional Notes

- none at this time

### DIFF
--- a/.changeset/smooth-ants-live.md
+++ b/.changeset/smooth-ants-live.md
@@ -1,0 +1,5 @@
+---
+"task-master-ai": minor
+---
+
+Added option for the AI to determine the number of tasks required based entirely on complexity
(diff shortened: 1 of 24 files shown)


geoh:
Totally can go inside the base config file. I nearly put it there originally, but didn't because everything in that file is generic across the whole system, whereas these settings only applied to Claude Code. I'll update it to use the default config file 👍

geoh:
I have reworked everything to use the default config file and incorporated the ability to define command specific overrides.

The code is much neater now.

I decided on 'claudeCode' instead of 'claude-code' for the config property so that we could use easily use optional chaining.

Crunchyman-ralph (approved):
lgtm

Issue #862, opened 2025-06-24
bug: expand command does not work with WSL / Claude Code because of using home path instead of project path


LucaL1fe:
### Description

Hey, it seems task-master has a problem to get the correct path using expand wiht WSL and Claude Code. 
I get:

"Error expanding task"
"Invalid tasks data"

To make our life easier I will make a unlisted video from the issue, with s step by step guide to reproduce:

https://youtu.be/2RHCKuwHX24

Crunchyman-ralph:
@LucaL1fe thank you so much for the video, its an awesome explanation of whats going on, let me fix that for you on next patch release/release!

LucaL1fe:
Wow! 5 Minutes?! Thank you very much for the quick response and fixing the issue🙏

Cluster444:
Hey @LucaL1fe if you `git init` or `touch package.json` in that directory, does that allow you to expand tasks until we can fix this properly?

LucaL1fe:
@Cluster444 
tried touch package.json, but seems to have problems with Claude Code
![Image](https://github.com/user-attachments/assets/01b4bd1e-a7db-4f9d-a2c7-e8adb28d04d7)

Cluster444:
Hmmm, that seems like a separate issue.

If you wouldn't mind, can you create a new dir. touch package.json inside it, then run tm init and the rest of the commands to see if it works then?

Testing a hunch

ben-vargas:
@LucaL1fe - Do you have a `package.json` or `.git` or `.taskmasterconfig` in a higher directory? Specifically in  `/mnt/c/Users/luca/`? Here is a brief analysis from claude code on the issue reported...

As a workaround, you might try the `--file` flag to explicitly point it to the correct tasks.json file, so `tm --expand-all --file:/mnt/c/Users/luca/drive/VS/breakfree/.taskmaster/tasks/tasks.json`

```
● Based on my analysis of the codebase, I can explain why task-master --expand-all is using the wrong directory path.

  The issue is in how task-master determines the project root directory. Looking at the path resolution logic in the codebase:

  1. findProjectRoot function (scripts/modules/utils.js:74-103) searches upward from the current working directory for project markers like:
    - package.json
    - .git
    - .taskmasterconfig (legacy config file)
  2. Project root detection priorities (from src/utils/path-utils.js:50-78):
    - .taskmaster directory
    - tasks.json files
    - Git markers (.git, .svn)
[...]

PR #877, opened 2025-06-26
Phase 1 refactoring path resolution.

Cluster444:
The current path resolution is implemented in a late binding fashion that is prone to circular dependencies and difficult to maintain. A full refactor is quite large so this is the first step of about 3-4 in total.

This PR introduces an initTaskMaster function that will handle all of the discovery and override logic, and produce a readonly object that can easily be passed down through the stack from the various entry points.

It includes some fairly comprehensive tests for the various scenarios that are likely to be encountered. And we may either want to amend for some edge cases, or it might be more prudent to simply deny some edge cases to maintain simplicity.

This phase only covers the CLI side, drilling down as far as the core APIs.

Some benefits of this also include being a bit more self documenting and ease of maintaining the .taskmaster directory, as this TaskMaster class is meant to be a 1:1 with that directory. So new features are easier to add.

This will also improve mocking abilities, as the inversion of control will allow this object to be a simple dependency injection, rather than requiring complex filesystem mocks in any tests that require pathing resolution.

### DIFF
--- a/scripts/modules/commands.js
+++ b/scripts/modules/commands.js
@@ -3737,7 +3737,11 @@ Examples:
 		$ task-master rules --${RULES_SETUP_ACTION}                  # Interactive setup to select rule profiles`
 		)
 		.action(async (action, profiles, options) => {
-			const projectDir = process.cwd();
+			const projectRoot = findProjectRoot();
+			if (!projectRoot) {
+				console.error(chalk.red('Error: Could not find project root.'));
+				process.exit(1);
+			}
 
 			/**
 			 * 'task-master rules --setup' action:
@@ -3784,7 +3788,7 @@ Examples:
 					const profileConfig = getRulesProfile(profile);
 
 					const addResult = convertAllRulesToProfileRules(
-						projectDir,
+						projectRoot,
 						profileConfig
 					);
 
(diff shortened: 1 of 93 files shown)


Cluster444:
ping #862

scripts/modules/commands.js
```diff
@@ -879,50 +875,9 @@ function registerCommands(programInstance) {
...
-						console.log(
-							chalk.blue(`Using default PRD file path: ${defaultPrdPath}`)
-						);
-						if (!(await confirmOverwriteIfNeeded())) return;
-
-						console.log(chalk.blue(`Generating ${numTasks} tasks...`));
-						spinner = ora('Parsing PRD and generating tasks...\n').start();
-						await parsePRD(defaultPrdPath, outputPath, numTasks, {
-							append: useAppend, // Changed key from useAppend to append
-							force: useForce, // Changed key from useForce to force
```
Crunchyman-ralph, line 900: curious as to why we removed this whole part.

Cluster444:
Yes that was my plan. I wanted to put this up to see how it looked.

Doing this with the mcp actually has the biggest impact because it removes a lot of error handling.

I'll get a new PR ready for the mcp side together shortly!

scripts/modules/commands.js
Cluster444: So this code should be effectively unreachable now give that we not longer deal with default vs non-default prd paths at this point, that is resolved closer to the command entry point. So it should mostly be unreachable duplication.

That said, we probably want to lift that Usage statement up and make it part of the error response at the entry point where this fails.

Crunchyman-ralph (approved):
lgtm, lets merge! Just needs to pass CI

PR #882, opened 2025-06-26
feat: Centralize AI prompts into JSON templates

joedanz:
### Overview
This PR refactors all hardcoded AI prompts throughout the codebase into a centralized prompt management system. All prompts are now stored as JSON templates in `/src/prompts/`, making them easier to maintain, update, and version control separately from the business logic.

### What Changed
- **Added PromptManager** (`scripts/modules/prompt-manager.js`): A new module that handles loading, caching, and rendering prompt templates with variable substitution and variant selection
- **Created 8 prompt templates** in `/src/prompts/`:
  - `parse-prd.json` - PRD parsing prompts
  - `expand-task.json` - Task expansion prompts  
  - `add-task.json` - Task creation prompts
  - `update-tasks.json` - Bulk task update prompts
  - `update-task.json` - Single task update prompts
  - `update-subtask.json` - Subtask update prompts
  - `analyze-complexity.json` - Complexity analysis prompts
  - `research.json` - Research query prompts
- **Migrated all 8 AI-powered modules** to use the centralized system
- **Added comprehensive documentation** in `/src/prompts/README.md`

### Why This Change
Previously, AI prompts were hardcoded as multi-line strings scattered throughout the codebase. This made them:
- Difficult to find and modify
- Hard to track changes over time
- Impossible to update without touching business logic
- Challenging to maintain consistency across features

### Benefits
- **Separation of Concerns**: Prompts are no longer embedded in business logic
- **Easy Maintenance**: Update prompts without touching code or risking logic bugs
- **Better Version Control**: Track prompt evolution separately from code changes
- **Consistent Structure**: All prompts follow the same format and patterns
- **Future Ready**: Foundation for A/B testing, hot reloading, and user customization

### Technical Details
The system supports:
- Variable substitution using Handlebars-like syntax (`{{variable}}`)
- Conditional content (`{{#if condition}}...{{/if}}`)
- Array iteration (`{{#each array}}...{{/each}}`)
- JSON serialization for complex objects (`{{{json object}}}`)
- Variant selection based on runtime conditions (e.g., research mode, append mode)

### Testing
- All existing tests pass without modification
- The external behavior remains identical - this is purely an internal refactoring
- Prompts generate the exact same output as before

### Breaking Changes
None. This is an internal refactoring with no changes to external APIs or behavior.

PR #884, opened 2025-06-26
Feat: Added automatic determination of task number based on complexity

geoh:
## Description

This PR allows the hard definition of the number of tasks / subtasks generated by 'parse-prd' and 'expand-task' commands to be relaxed by using '0' value in the config or command parameter

## Type of Change

- [ ] Bug fix
- [x] New feature
- [ ] Breaking change
- [ ] Documentation update

## Testing

- [ ] All existing tests pass
- [ ] I have added tests for new functionality

## Changeset

## Additional Notes

- Not entirely sure about adding conditional logic to the prompts, but there is already some there...
- This PR is not compatible with #882 - which is an overhaul of the prompting system

### DIFF
--- a/scripts/modules/commands.js
+++ b/scripts/modules/commands.js
@@ -69,7 +69,9 @@ import {
 	ConfigurationError,
 	isConfigFilePresent,
 	getAvailableModels,
-	getBaseUrlForRole
+	getBaseUrlForRole,
+	getDefaultNumTasks,
+	getDefaultSubtasks
 } from './config-manager.js';
 
 import { CUSTOM_PROVIDERS } from '../../src/constants/providers.js';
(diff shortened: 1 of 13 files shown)


scripts/modules/commands.js
```diff
@@ -69,7 +69,9 @@ import {
 	ConfigurationError,
 	isConfigFilePresent,
 	getAvailableModels,
-	getBaseUrlForRole
+	getBaseUrlForRole,
+	getDefaultNumTasks,
+	getDefaultSubtasks
```
Crunchyman-ralph, line 74: why are we importing `getDefaultSubtasks` if we're not using it

Crunchyman-ralph (approved):
lgtm

PR #916, opened 2025-07-03
feat: Introduce modern E2E test suite for Task Master AI

Crunchyman-ralph:


Crunchyman-ralph:
Too slow, need to break down into:
- unit tests
- integration tests
- e2e tests (fewer because you did a good job with integration tests) [do the first ones first to catch issues earlier so that your e2e tests while they take longer, ideally pass]

PR #941, opened 2025-07-08
Typescript Migration: Continued

Cluster444:
This brings together the work from @neno-is-ooo and gets the last bits necessary to pull in a PR before beginning the piece-meal migration.


Crunchyman-ralph:
Ended up doing a typescript migration:
#1190

PR #951, opened 2025-07-10
refactor(commands): Update tasks path retrieval to use taskMaster.getTasksPath() for consistency

Crunchyman-ralph:


Crunchyman-ralph:
Already implemented

PR #956, opened 2025-07-11
fix: more regression bugs

Crunchyman-ralph:


src/task-master.js
```diff
@@ -218,7 +218,13 @@ export function initTaskMaster(overrides = {}) {
 		);
 	}
 
-	// Remaining paths - only resolve if key exists in overrides
+	// Always set default paths first
+	// These can be overridden below if needed
+	paths.configPath = path.join(paths.projectRoot, TASKMASTER_CONFIG_FILE);
+	paths.statePath = path.join(paths.taskMasterDir || path.join(paths.projectRoot, TASKMASTER_DIR), 'state.json');
+	paths.tasksPath = path.join(paths.projectRoot, TASKMASTER_TASKS_FILE);
```
ben-vargas: o3 flagged that this will break tests, but that's probably known from CI - looking at commands.js while we will never return nul to some of the checks there, I don't think this will break anything. Probably safe to proceed assuming failing tests are adjusted.

scripts/modules/task-manager/research.js
```diff
@@ -205,12 +205,10 @@ async function performResearch(
...
 		};
 
-		// Select variant based on detail level
-		const variantKey = detailLevel; // 'low', 'medium', or 'high'
+		// Load prompts - the research template handles detail level internally
 		const { systemPrompt, userPrompt } = await promptManager.loadPrompt(
 			'research',
-			promptParams,
-			variantKey
+			promptParams
```
ben-vargas, line 211: Nice, pattern/behavior consistency with analyze-complexity.js

scripts/modules/task-manager/expand-task.js
```diff
@@ -469,7 +469,7 @@ async function expandTask(
 			complexityReasoningContext: complexityReasoningContext,
 			gatheredContext: gatheredContext,
 			useResearch: useResearch,
-			expansionPrompt: taskAnalysis?.expansionPrompt || null
+			expansionPrompt: taskAnalysis?.expansionPrompt || undefined
```
ben-vargas, line 472: Asked AI to scan for downstream impacts of undefined rather than null... all downstream evaluations are truthy/falsy so this appears good to go.

PR #1040, opened 2025-07-24
Feat/add-kilocode-rules

DomVidja:
Description
Add Kilo Code profile integration with custom modes and MCP configuration. This PR introduces support for Kilo Code AI assistant, including profile management, custom mode configurations, and proper asset handling for .kilocodemodes and mode-specific rules.

Type of Change
[ ] Bug fix
[x] New feature
[ ] Breaking change
[ ] Documentation update
Testing
[x] I have tested this locally
[x] All existing tests pass (1007/1007)
[x] I have added tests for new functionality
Changeset
[x] I have created a changeset (or this change doesn't need one)
Additional Notes
This integration adds:

Kilo Code profile with proper lifecycle management (add/remove/post-convert)
Support for custom modes through .kilocodemodes configuration
Asset copying for mode-specific rules in .kilo/rules-{mode}/ directories
Integration with the existing profile system using ROO_STYLE tool mappings
Proper cleanup when removing the profile
The implementation follows the established patterns in the codebase and maintains compatibility with existing functionality. All tests pass and code formatting has been applied.


assets/kilocode/.kilo/.kilocodemodes
Crunchyman-ralph, line 1: Can we try to use our existing rules, or does kilo code need its own wording ? See how we do it with roo or cline

src/constants/profiles.js
```diff
@@ -59,6 +61,20 @@ export const ROO_MODES = [
...
+ * @description Available Kilo Code modes for rule generation
+ */
+export const KILO_MODES = [
+	'architect',
+	'ask',
+	'orchestrator',
+	'code',
+	'debug',
+	'test'
+];
```
Crunchyman-ralph: why do we need this ?

src/profiles/kilo.js
Crunchyman-ralph, line 1: make sure we're leveraging our existing functions, look at roo.js or cline.js

Crunchyman-ralph (changes requested):
small changes, but would love to get this out, seems exciting. wondering if we can consolidate our rules between all the different IDEs

assets/kilocode/.kilo/.kilocodemodes
DomVidja: Consolidating rules would be great! There is a new IDE every other week these days... but because they all have slightly different quirks, that might not be completely trivial. 

So I reviewed the roo implementation and essentially tried to replicate that exact implementation (same rules, same modes, same functions) but with the kilo / kilo code branding. I think it should be good to go now

src/constants/profiles.js
DomVidja: We technically could reuse the const ROO_MODES. I just wanted to be consistent and create the const KILO_MODES. It should be fine to remove this and use ROO_MODES variable

Crunchyman-ralph: could we just do KILO_MODES = ROO_MODES ? if you wanna keep the KILO_MODES name ?

assets/kilocode/.kilo/.kilocodemodes
Crunchyman-ralph: @DomVidja I don't see any new commits, feel free to commit and then remove the assets files related to kilo code when you've done that.

What I suggest is to re-use roo code rules, and just do a regex replace inside the profile file of kilo, as mentioned, look at how other profiles do it. @joedanz can input his expertise here

joedanz:
Thanks for the contribution @DomVidja !  We've actually got a profile system setup where you should essentially be able to do this just by adding a new kilocode.js under /src/profiles and making sure its references under /src/profiles.

Is this essentially identical to Roo profile with with "Kilo" instead?

src/constants/profiles.js
DomVidja: I actually ended up removing the KILO_MODES constant and instead importing ROO_MODES directly

DomVidja:
> Thanks for the contribution @DomVidja ! We've actually got a profile system setup where you should essentially be able to do this just by adding a new kilocode.js under /src/profiles and making sure its references under /src/profiles.
> 
> Is this essentially identical to Roo profile with with "Kilo" instead?

Hi @joedanz - it's my pleasure. I love TaskMaster and I'm happy to contribute. 

You are absolutely correct. This should be essentially identical to Roo profile but with Kilo branding instead, given that Kilo is originally a fork of Roo. I've made some changes to the code following @Crunchyman-ralph advice and tried to leverage the existing Roo functions for less duplication. I committed now, so it should be ready for a review.

PR #1062, opened 2025-07-31
fix: changed the GEMINI api key name to GOOGLE api key name

shalevo13:
# What type of PR is this?

 - [x] 🐛 Bug fix
 - [ ] ✨ Feature
 - [ ] 🔌 Integration
 - [ ] 📝 Docs
 - [ ] 🧹 Refactor
 - [ ] Other:
## Description
This PR fixes an issue with the Gemini CLI provider where it was looking for the environment variable `GEMINI_API_KEY` instead of the standard `GOOGLE_API_KEY` used by all other Google-related providers in the codebase.

**Expected result:**
Env variable inside user confige can now start with GOOGLE, and not GEMINI, as the .env.example says.

## Contributor Checklist

- [x] Created changeset: `npm run changeset`
- [x] Tests pass: `npm test`
- [x] Format check passes: `npm run format-check` (or `npm run format` to fix)
- [x] Addressed CodeRabbit comments (if any)
- [x] Linked related issues (if any)
- [x] Manually tested the changes

## Changelog Entry
"Changed GEMINI_API_KEY to GOOGLE_API_KEY"

### For Maintainers

- [ ] PR title follows conventional commits
- [ ] Target branch correct
- [ ] Labels added
- [ ] Milestone assigned (if applicable)

### DIFF
--- a/docs/providers/gemini-cli.md
+++ b/docs/providers/gemini-cli.md
@@ -37,7 +37,7 @@ For OAuth use, select `Login with Google` - This will open a browser window for 
 While the primary use case is OAuth authentication, you can also use an API key if needed:
 
 ```bash
-export GEMINI_API_KEY="your-gemini-api-key"
+export GOOGLE_API_KEY="your-google-api-key"
 ```
 
 **Note:** If you want to use API keys, consider using the standard `google` provider instead, as `gemini-cli` is specifically designed for OAuth/subscription users.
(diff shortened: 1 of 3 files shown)


shalevo13:
> May I ask why you're trying to do this ? is it a big issue to have it GEMINI_API_KEY ?

Not that big of an issue.

The .env.example file does not contain the GEMINI_API_KEY naming, thus leaving you with the impression that the GOOGLE_API_KEY is the one to insert your API key, if you are using the gemini 2.5 flash / pro models.
This is a mistake, since the code looks for the GEMINI_API_KEY instead.
Finally, the error you are left with is: 
[error] {"message":"Error parsing PRD: Gemini CLI API error during object generation: Failed to initialize Gemini model: TypeError: config.isBrowserLaunchSuppressed is not a function"}

Crunchyman-ralph:
@shalevo13 I think its too big of a behavioral change for users to have to move from GEMINI_API_KEY to GOOGLE_API_KEY for no real value add.

I'm afraid I don't think I'll be able to merge this PR, although you do raise a good point, maybe we should add GEMINI_API_KEY into .env.example, but the thing is that most users just login to gemini cli and don't use a gemini_api_key which is probably why this issue didn't come up, I think instead of adding as many env variables to the .env.example, we should just document it better in our docs: docs.task-master.dev

PR #1462, opened 2025-11-30
fix: add terminal-link to root dependencies for global install

eyaltoledano:
## Problem

When users install `task-master-ai@rc` globally and run `tm list`, they get:

```
Error [ERR_MODULE_NOT_FOUND]: Cannot find package 'terminal-link' imported from .../dist/dependency-manager-DIOZ_B7q.js
```

## Root Cause

The `terminal-link` package is used by `apps/cli/src/ui/formatters/link-formatters.ts` but was only listed as a dependency in `apps/cli/package.json`, not in the root `package.json`.

Since the bundler (tsdown) keeps npm dependencies external (only bundling `@tm/*` workspace packages), when users install `task-master-ai` globally, only the dependencies from the root `package.json` are installed — so `terminal-link` was missing.

## Fix

Added `"terminal-link": "^5.0.0"` to root `package.json` dependencies.

## Summary by CodeRabbit

* **Chores**
  * Updated project dependencies

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -1,12 +1,12 @@
 {
 	"name": "task-master-ai",
-	"version": "0.36.0-rc.0",
+	"version": "0.36.0-rc.1",
 	"lockfileVersion": 3,
 	"requires": true,
 	"packages": {
 		"": {
 			"name": "task-master-ai",
-			"version": "0.36.0-rc.0",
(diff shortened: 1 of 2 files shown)


Crunchyman-ralph (approved):
lgtm

PR #1475, opened 2025-12-02
fix(auth): enforce org selection post-login and remove duplicate messages

eyaltoledano:
## Summary
Enforces organization selection after login across all CLI commands that require authentication, and removes duplicate 'Authentication successful!' messages.

## Changes
- Add `ensureOrgSelected` to auth-guard.ts for all auth flows (existing & new sessions)
- Add `ensureOrgSelected` to parse-prd flow in commands.js (after authentication)
- Add `ensureOrgSelected` to init.js cloud storage flow (using shared utility)
- Remove duplicate 'Authentication successful!' boxen messages
- Export `authenticateWithBrowserMFA` and `ensureOrgSelected` from `@tm/cli` utils
- init.js now uses shared `authenticateWithBrowserMFA` instead of custom OAuth flow
- auth-guard.ts now checks `context.orgId` before making API calls (optimization)

## Fixes
- Org selection now happens after login in parse-prd, init, and export commands
- Single source of truth for browser auth with MFA support
- Removed redundant auth UI code from init.js

## Known Issues (needs investigation)
- `tm init` and `tm export` may hang after completion
- Root cause: Supabase client's `autoRefreshToken: true` keeps Node.js event loop alive
- `tm auth login` works fine (uses different flow via `setupContextInteractive`)
- Proper fix: add cleanup method to SupabaseAuthClient to stop auto-refresh timer

## Testing
- [ ] `tm auth login` - works, sets org properly
- [ ] `tm parse-prd` - org selection works (may auto-select if only one org)
- [ ] `tm export` - auth + org works, but may hang after completion
- [ ] `tm init` (cloud) - auth + org works, but may hang after completion

## Summary by CodeRabbit

* **New Features**
  * Added organization selection requirement during authentication workflows to ensure proper context configuration.

* **Bug Fixes**
  * Improved error handling for team invitation responses to support multiple API response formats and better detect "already member" scenarios.

* **Refactor**
  * Streamlined authentication flow with centralized multi-factor authentication handling.
  * Updated success messaging to log user identity instead of displaying a separate success notification.

### DIFF
--- a/apps/cli/src/utils/auth-guard.ts
+++ b/apps/cli/src/utils/auth-guard.ts
@@ -76,7 +76,7 @@ export async function ensureAuthenticated(
 	if (hasSession) {
 		// Only get AuthManager when we need to check org selection
 		const authManager = AuthManager.getInstance();
-		
+
 		// Check if org is already selected (quick check before any API calls)
 		const context = authManager.getContext();
 		if (context?.orgId) {


apps/cli/src/commands/export.command.ts
```diff
@@ -334,6 +335,25 @@ export class ExportCommand extends Command {
 				return;
 			}
 
+			// Force org selection after tag selection
+			// User can choose which org to export to, with current org pre-selected
+			const authManager = (this.taskMasterCore!.auth as any).authManager;
```
Crunchyman-ralph: maybe just import from tm-core what you need directly ?

Crunchyman-ralph (approved):
lgtm

PR #1566, opened 2026-01-07
fix: Add cross-process file locking to prevent race conditions

### LINKED ISSUE #1567: Race condition when multiple Claude Code windows write to tasks.json
## Problem

When multiple Claude Code windows (separate processes) access Task Master simultaneously, the JSON file can get corrupted or work that Task Master claims to have saved is lost.

## Root Cause

The `readJSON()` and `writeJSON()` functions in `scripts/modules/utils.js` had no file locking mechanism:
- Multiple processes could read the same file simultaneously
- Each process would make changes to its copy
- When writing, later writes would overwrite earlier ones, losing updates

Additionally, the `_rawTaggedData` snapshot pattern caused stale data issues - a snapshot taken at read time could become outdated by the time the write occurred.

## Impact

- Data loss when multiple Claude Code sessions work on the same project
- Corrupted tasks.json file
- Lost task updates, status changes, and subtask modifications

## Solution

Implemented in PR #1566:
- Cross-process file locking using `proper-lockfile` (async) and custom lock files with exclusive `wx` flag (sync)
- Atomic writes using temp file + rename pattern
- Re-read file inside lock to get current state
- Stale lock detection and automatic cleanup (10-second timeout)

## Reproduction Steps

1. Open multiple Claude Code windows on the same project
2. Have both windows make Task Master operations simultaneously (e.g., set-status, update-subtask)
3. Observe that some changes are lost or file becomes corrupted

bjcoombs:
## Summary

Fixes #1567

This PR addresses a race condition where multiple Claude Code windows (separate processes) accessing Task Master simultaneously could corrupt the JSON file or cause saved work to be lost.

### File Locking
- Added `withFileLock` (async) and `withFileLockSync` (sync) utilities for cross-process file locking
- Uses `proper-lockfile` for async operations with configurable retries and stale detection
- Custom lock file approach with exclusive `wx` flag for sync operations

### Atomic Writes
- Implemented temp file + rename pattern to prevent partial writes
- Re-read file inside lock to get current state, preventing lost updates from stale snapshots

### Safety Features
- Stale lock detection and automatic cleanup (10-second timeout)
- Race-condition-safe implementation using atomic file creation (`wx` flag)
- Replaced busy-wait with `Atomics.wait()` for proper blocking sleep

### Bug Fixes (discovered during review)
- Fixed metadata structure bug where metadata was incorrectly spread at tag level
- Fixed cleanup logic that was discarding `created` timestamp
- Fixed retry count inconsistency between async and sync locking

## Test Plan

- [x] Added comprehensive test suite for file locking (`tests/unit/file-locking.test.js`)
- [x] Tests cover: lock acquisition, release, cleanup on error, stale detection, atomic writes, concurrent simulation
- [x] All 13 tests pass locally and in CI
- [x] Existing test suite passes

---

> [!NOTE]
> Strengthens file write safety across processes to avoid lost updates and partial writes.
> 
> - Introduces `withFileLock`/`withFileLockSync` in `scripts/modules/utils.js` with retries, stale-lock cleanup (10s), and atomic temp-file + rename; exported for reuse
> - Refactors `FileOperations` to use `proper-lockfile` + `steno` for atomic writes; adds `modifyJson` that re-reads inside the lock, `ensureFileExists` (atomic create with `wx`), writer caching, and `cleanup`
> - Adds tests: Vitest for `FileOperations` and Jest suite simulating concurrent/multi-process writes; verifies serialization and no leftover lock/temp files
> - Adds deps `proper-lockfile` and `@types/proper-lockfile`; updates package metadata and adds a changeset entry
> 
> <sup>Written by [Cursor Bugbot](https://cursor.com/dashboard?tab=bugbot) for commit 0b05a35b3417f18a58f77fd2f14cdbf8cc1623ce. This will update automatically on new commits. Configure [here](https://cursor.com/dashboard?tab=bugbot).</sup>

## Summary by CodeRabbit

* **Bug Fixes**
  * Prevented data loss and race conditions when multiple windows access shared task files by adding cross-process locking, atomic writes, and stale-lock timeout/cleanup.

* **New Features**
  * Exposed reusable file-locking utilities for serialized read-modify-write operations and safer atomic file updates.

* **Tests**
  * Added comprehensive tests covering locking, atomic writes, concurrent multi-process writes, and cleanup behavior.

* **Documentation**
  * Added changelog entry describing the fix and behavior.

### DIFF
--- a/packages/tm-core/src/modules/storage/adapters/file-storage/file-operations.ts
+++ b/packages/tm-core/src/modules/storage/adapters/file-storage/file-operations.ts
@@ -27,7 +27,10 @@ const LOCK_OPTIONS = {
 };
 
 /**
- * Handles atomic file operations with cross-process locking mechanism
+ * Handles atomic file operations with cross-process locking mechanism.
+ *
+ * Writers are cached for reuse. Call {@link cleanup} when disposing of
+ * long-lived instances to prevent memory leaks.
  */
 export class FileOperations {
 	/** Map of file paths to steno Writers for reuse */
@@ -89,8 +92,14 @@ export class FileOperations {
 			if (release) {
 				try {
 					await release();
-				} catch {
-					// Ignore release errors - lock may have been released already
+				} catch (err: any) {
+					// Log but don't throw - lock may have been released already
+					// Other errors should be visible for debugging
+					if (process.env.DEBUG || process.env.TASKMASTER_DEBUG === 'true') {
+						console.warn(
+							`[WARN] Lock release warning for ${filePath}: ${err.message}`
+						);
+					}
 				}
 			}
 		}
@@ -121,9 +130,28 @@ export class FileOperations {
 			try {
 				const content = await fs.readFile(filePath, 'utf-8');
 				currentData = JSON.parse(content);
-			} catch {
-				// File empty or invalid - start with empty object
-				currentData = {} as T;
+			} catch (err: any) {
+				// Distinguish between expected empty/new files and actual corruption
+				if (err.code === 'ENOENT') {
+					// File doesn't exist yet - start fresh
+					currentData = {} as T;
+				} else if (err instanceof SyntaxError) {
+					// Check if it's just an empty file (our ensureFileExists writes '{}')
+					const content = await fs.readFile(filePath, 'utf-8').catch(() => '');
+					if (content.trim() === '' || content.trim() === '{}') {
+						currentData = {} as T;
+					} else {
+						// Actual JSON corruption - this is a serious error
+						throw new Error(
+							`Corrupted JSON in ${filePath}: ${err.message}. File contains: ${content.substring(0, 100)}...`
+						);
+					}
+				} else {
+					// Other errors (permission, I/O) should be surfaced
+					throw new Error(
+						`Failed to read ${filePath} for modification: ${err.message}`
+					);
+				}
 			}
 
 			// Apply modification
@@ -137,8 +165,14 @@ export class FileOperations {
 			if (release) {
 				try {
 					await release();
-				} catch {
-					// Ignore release errors - lock may have been released already
+				} catch (err: any) {
+					// Log but don't throw - lock may have been released already
+					// Other errors should be visible for debugging
+					if (process.env.DEBUG || process.env.TASKMASTER_DEBUG === 'true') {
+						console.warn(
+							`[WARN] Lock release warning for ${filePath}: ${err.message}`
+						);
+					}
 				}
 			}
 		}
--- a/packages/tm-core/src/modules/storage/adapters/file-storage/file-operations.spec.ts
+++ b/packages/tm-core/src/modules/storage/adapters/file-storage/file-operations.spec.ts
@@ -1,0 +1,264 @@
+/**
+ * Tests for FileOperations class
+ * Focuses on modifyJson and cross-process locking functionality
+ */
+
+import { describe, it, expect, beforeEach, afterEach } from 'vitest';
+import fs from 'node:fs/promises';
+import fsSync from 'node:fs';
+import path from 'node:path';
+import os from 'node:os';
+import { FileOperations } from './file-operations.js';
+
+describe('FileOperations', () => {
+	let tempDir: string;
+	let testFilePath: string;
+	let fileOps: FileOperations;
+
+	beforeEach(async () => {
+		// Create a temp directory for each test
+		tempDir = fsSync.mkdtempSync(path.join(os.tmpdir(), 'tm-core-test-'));
+		testFilePath = path.join(tempDir, 'test.json');
+		fileOps = new FileOperations();
+	});
+
+	afterEach(async () => {
+		// Clean up
+		await fileOps.cleanup();
+		if (tempDir && fsSync.existsSync(tempDir)) {
+			fsSync.rmSync(tempDir, { recursive: true, force: true });
+		}
+	});
+
+	describe('modifyJson', () => {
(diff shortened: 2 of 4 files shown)


Crunchyman-ralph (changes requested):
nice initiative, we use steno for fire locking, maybe try using that instead ?
[steno](https://www.npmjs.com/package/steno)

see an example of how we use it in:
`packages/tm-core/src/modules/workflow/managers/workflow-state-manager.ts`

Also, make sure that any new code is written in typescript inside packages/* (it can be added in tm-bridge if you can't put it in tm-core) so that we are gradually transitioning into a typescript codebase and moving away from javascript for type safety

bjcoombs:
Thanks for the feedback @Crunchyman-ralph! I've updated the TypeScript implementation to use `steno` for atomic writes following the same pattern as `workflow-state-manager.ts`.

**Changes made:**
- Import and use `steno`'s `Writer` class for atomic file writes
- Cache Writer instances per file path for reuse (same pattern as workflow-state-manager)
- Removed custom `performAtomicWrite` method

**Note on approach:**
I kept `proper-lockfile` for cross-process locking because:
- `steno` handles atomic writes and queues concurrent writes within a **single process**
- `proper-lockfile` prevents lost updates when **multiple processes** (e.g., multiple Claude Code windows) access the same file simultaneously

The original bug report was about data loss when multiple Claude Code windows write to tasks.json at the same time. Since these are separate processes, we need cross-process locking to prevent the "lost update" problem where:
1. Process A reads tasks.json
2. Process B reads tasks.json  
3. Process A writes changes (adds task X)
4. Process B writes changes (overwrites A's changes, losing task X)

The combination of steno (for atomic writes) + proper-lockfile (for cross-process locking) + re-read inside lock (to get current state) provides comprehensive protection.

bjcoombs:
## Regarding CodeRabbit's Critical Feedback

Thanks for the thorough analysis! Let me address both points:

### 1. FileStorage.close() not being called

This is a pre-existing issue in the codebase that exists independently of this PR. My change actually adds functionality to `cleanup()` that makes it useful (clearing the Writers Map) - before it was a no-op. I'd recommend addressing the broader "call close() on FileStorage instances" issue as a follow-up PR since it touches many files across the codebase.

### 2. Using `modifyJson()` for read-modify-write operations

This is a great catch! I added the `modifyJson()` method specifically to solve the lost update problem with the read-inside-lock pattern. However, there are two code paths to consider:

1. **Legacy JavaScript** (`scripts/modules/utils.js`): This is what's currently used for CLI operations. I've updated `writeJSON()` there with file locking, but the pattern of read-outside, modify, write is the same. The lock prevents concurrent writes but doesn't prevent stale reads.

2. **TypeScript tm-core** (`file-storage.ts`): This has `modifyJson()` available but isn't using it yet. Refactoring all methods to use `modifyJson()` would be a significant change.

**Proposal for this PR**: Focus on the immediate bug fix (prevent file corruption via locking + atomic writes). The enhancement to use `modifyJson()` everywhere would be better as a follow-up PR to avoid scope creep and potential regressions.

The current fix still provides significant improvement:
- Prevents partial writes (atomic via steno)
- Prevents concurrent write collisions (cross-process locking)
- The `modifyJson()` method is available for gradual migration

@Crunchyman-ralph what are your thoughts on scope? Should we:
A) Keep this PR focused on the immediate fix and address `modifyJson()` adoption as follow-up?
B) Expand this PR to migrate file-storage.ts methods to use `modifyJson()`?

bjcoombs:
## Follow-up Issue Created

I've created issue #1568 to track the broader work of adopting `modifyJson()` pattern throughout the codebase:

**https://github.com/eyaltoledano/claude-task-master/issues/1568**

This PR (#1566) focuses on the immediate fix:
- ✅ Cross-process file locking (prevents concurrent write collisions)
- ✅ Atomic writes via steno (prevents partial/corrupt files)
- ✅ `modifyJson()` method available for future use

Issue #1568 covers the follow-up work:
- Migrate file-storage.ts methods to use `modifyJson()`
- Ensure FileStorage.close() is called throughout codebase
- Consider legacy JavaScript code migration

This keeps the current PR focused and reduces risk of regressions.

Crunchyman-ralph:
@bjcoombs lets fix the modifyJson issue in a separate PR, but maybe merge both at the same time ?

packages/tm-core/src/modules/storage/adapters/file-storage/file-operations.ts
```diff
@@ -1,16 +1,49 @@
...
+ *
+ * Uses steno for atomic writes (same pattern as workflow-state-manager.ts)
+ * and proper-lockfile for cross-process locking to prevent lost updates.
  */
 
 import { constants } from 'node:fs';
 import fs from 'node:fs/promises';
+import path from 'node:path';
+import lockfile from 'proper-lockfile';
+import { Writer } from 'steno';
```
Crunchyman-ralph, line 12: I'm not sure, but I think steno handles lockfile ? meaning proper-lockfile might not be needed ? If it is, keep it, but just challenging you on that.

bjcoombs:
@Crunchyman-ralph I've created the separate PR as requested:

**PR #1569**: [fix: adopt modifyJson pattern for atomic read-modify-write operations](https://github.com/eyaltoledano/claude-task-master/pull/1569)

This PR depends on #1566 (it includes those changes via merge) and can be merged together as you suggested.

**Summary of both PRs:**
- **#1566**: Adds cross-process file locking + atomic writes via steno + `modifyJson()` method
- **#1569**: Migrates file-storage.ts methods to use `modifyJson()` pattern

PR #1696, opened 2026-04-28
fix: disable AI content recording and PII collection in Sentry by default

### LINKED ISSUE #1681: WARNING: DEFAULT Sentry telemetry records 100% of full AI prompts/responses with sendDefaultPii:true —   opt-out is fragile
Summary

  **The default telemetry configuration captures 100% of all AI prompts and responses
  (including PRD content, task descriptions, and AI-generated code) and sends them to a
   Sentry instance with PII collection enabled. The opt-out mechanism is fragile and
  can silently fail.**

  There is a dangerous pattern in the space:
  1. Ship a popular open-source tool (26K stars)
  2. Default telemetry ON with sendDefaultPii: true + recordInputs: true +
  recordOutputs: true at 100% sample rate
  3. Make the opt-out check fragile (module-level init with no projectRoot, silently
  falls back to ON)
  4. Wrap the entire MCP protocol with Sentry instrumentation on top of that
  5. Capture every PRD, task description, and AI-generated code response from thousands
   of developers
  6. Call it "anonymous telemetry" even though sendDefaultPii: true is literally the
  opposite

  What's recorded (when anonymousTelemetry: true, which is the default)

  From sentry-CZp3yiwv.js:

  r.init({
    dsn: "https://ce8c03ca...@o4510099843776512.ingest.us.sentry.io/...",
    integrations: [
      r.vercelAIIntegration({ recordInputs: true, recordOutputs: true }),
      r.zodErrorsIntegration()
    ],
    tracesSampleRate: 1,      // 100% of operations
    sendDefaultPii: true,     // PII collection ON
  })

  vercelAIIntegration({recordInputs: true, recordOutputs: true}) monkey-patches every
  Vercel AI SDK method (generateText, streamText, generateObject, streamObject) via
  OpenTelemetry. This means the full text of every prompt sent to any AI provider and
  every response received is captured as span attributes and sent to your Sentry
  instance.

  For a developer tool that processes PRDs, codebases, and task descriptions, this
  means you are collecting:
  - Full PRD/requirements document text
  - Task descriptions and implementation details
  - AI-generated code and analysis
  - All of the above at 100% sample rate with PII enabled

  The opt-out is fragile

  The config check in the sentry init function relies on reading
  .taskmaster/config.json from projectRoot. However, ie() is called at module load time
   with no arguments:

  // mcp-server.js (module level)
  ie();  // no projectRoot passed

  Inside the init function:
  function o(n={}) {
    // n.projectRoot is undefined
    try {
      if (!t(n.projectRoot)) { return; }  // t() resolves config from CWD
    } catch {} finally { e(false) }
    // If config resolution fails silently, falls through to Sentry.init()
  }

  If the CWD doesn't contain .taskmaster/config.json when the MCP server process starts
   (which depends on how the MCP client spawns it), the check fails silently and Sentry
   initializes with full recording.

  Additionally, wrapMcpServerWithSentry wraps the entire MCP protocol regardless of the
   config check.

  Recommendations

  1. Change the default to anonymousTelemetry: false (opt-in, not opt-out)
  2. Remove sendDefaultPii: true — crash reporting doesn't need PII
  3. Remove recordInputs: true, recordOutputs: true — error monitoring doesn't need
  full prompt/response capture
  4. Reduce tracesSampleRate from 1.0 — 100% is excessive for telemetry
  5. Pass projectRoot to ie() so the config check actually works reliably
  6. Don't wrap the MCP server with Sentry if telemetry is disabled
  7. Clearly disclose in the README what data is collected and where it goes

  Workaround

  Users can protect themselves by:
  1. Setting "anonymousTelemetry": false in .taskmaster/config.json
  2. Setting SENTRY_DSN="" in the MCP server env config (belt-and-suspenders)
  3. Pinning the package version instead of using @latest

octo-patch:
Fixes #1681

## Problem

The default Sentry telemetry configuration was recording **100% of AI prompts and responses** — including full PRD text, task descriptions, and AI-generated code — and sending them to Sentry with PII collection enabled:

```js
// Before:
Sentry.vercelAIIntegration({ recordInputs: true, recordOutputs: true })
sendDefaultPii: true
```

This meant every `generateText`, `streamText`, `generateObject`, and `streamObject` call was captured as a span attribute and sent to Sentry by default, even though the telemetry was described as "anonymous."

## Solution

Change the defaults to disable AI content recording and PII collection:

```js
// After:
Sentry.vercelAIIntegration({ recordInputs: false, recordOutputs: false })
sendDefaultPii: false
```

The same change is applied to `getAITelemetryConfig` for consistency.

Sentry still captures **errors and performance spans** (traces), which is the intended purpose of error tracking telemetry. Only the capture of full prompt/response content is disabled by default.

## Testing

- Reviewed Sentry init flow — error and trace capture still works with these settings
- `recordInputs`/`recordOutputs` only control whether AI span content is attached to traces; setting them to `false` does not break error tracking

---

> [!NOTE]
> **Medium Risk**
> Changes default Sentry telemetry to stop sending AI prompt/response content and PII, and reduces default trace sampling, which could slightly reduce observability if deployments relied on prior defaults.
> 
> **Overview**
> Tightens Sentry telemetry privacy defaults by **disabling AI prompt/response recording** (`recordInputs`/`recordOutputs` now default to `false`) both in `Sentry.vercelAIIntegration()` and the `getAITelemetryConfig()` used by AI SDK calls.
> 
> Also **opts out of PII collection by default** (`sendDefaultPii: false`) and lowers the default `tracesSampleRate` from `1.0` to `0.1`, while keeping error/performance telemetry enabled when configured.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit d747589dd7a3937855ca7e37dece9392ff31febe. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

## Summary by CodeRabbit

* **Bug Fixes**
  * AI prompt/response telemetry recording disabled by default to improve privacy.
  * PII collection disabled by default; explicit opt-in required.
  * Error tracking and performance monitoring remain active to maintain reliability.

* **Chores**
  * Reduced default performance tracing sample rate to lower production overhead.

### DIFF
--- a/.changeset/fix-sentry-pii-defaults.md
+++ b/.changeset/fix-sentry-pii-defaults.md
@@ -1,0 +1,5 @@
+---
+"task-master-ai": patch
+---
+
+Fix telemetry privacy defaults: AI prompt/response recording and PII collection are now disabled by default. Sentry telemetry still tracks errors and performance spans, but no longer sends AI prompt content, task descriptions, or PRD text to Sentry.
--- a/.changeset/fix-sentry-pii-defaults.md
+++ b/.changeset/fix-sentry-pii-defaults.md
@@ -2,4 +2,4 @@
 "task-master-ai": patch
 ---
 
-Fix telemetry privacy defaults: AI prompt/response recording and PII collection are now disabled by default. Sentry telemetry still tracks errors and performance spans, but no longer sends AI prompt content, task descriptions, or PRD text to Sentry.
+Fix telemetry privacy defaults: AI prompt/response recording and PII collection are now disabled by default. Sentry telemetry still tracks errors and performance spans, but no longer sends AI prompt content, task descriptions, or PRD text to Sentry. The default trace sample rate is also lowered from 1.0 to 0.1 to reduce production overhead.
--- a/src/telemetry/sentry.js
+++ b/src/telemetry/sentry.js
@@ -88,8 +88,10 @@ export function initializeSentry(options = {}) {
 				// Add Zod error tracking for better validation error reporting
 				Sentry.zodErrorsIntegration()
 			],
-			// Tracing must be enabled for AI monitoring to work
-			tracesSampleRate: options.tracesSampleRate ?? 1.0,
+			// Tracing must be enabled for AI monitoring to work.
+			// Default to 0.1 (10%) — Sentry's recommended production rate.
+			// Override via options.tracesSampleRate or SENTRY_TRACES_SAMPLE_RATE.
+			tracesSampleRate: options.tracesSampleRate ?? 0.1,
 			// Disabled to avoid capturing personally identifiable information.
 			sendDefaultPii: options.sendDefaultPii ?? false,
 			// Enable debug mode with SENTRY_DEBUG=true env var
@@ -102,7 +104,7 @@ export function initializeSentry(options = {}) {
 			console.log(
 				`  Environment: ${options.environment || process.env.NODE_ENV || 'production'}`
 			);
-			console.log(`  Traces Sample Rate: ${options.tracesSampleRate ?? 1.0}`);
+			console.log(`  Traces Sample Rate: ${options.tracesSampleRate ?? 0.1}`);
 		}
 	} catch (error) {
 		console.error(`Failed to initialize telemetry: ${error.message}`);
--- a/src/telemetry/sentry.js
+++ b/src/telemetry/sentry.js
@@ -79,16 +79,19 @@ export function initializeSentry(options = {}) {
 			environment: options.environment || process.env.NODE_ENV || 'production',
 			integrations: [
 				// Add the Vercel AI SDK integration for automatic AI operation tracking
+				// recordInputs/recordOutputs are disabled by default to avoid capturing
+				// sensitive content such as PRDs, task descriptions, and AI-generated code.
 				Sentry.vercelAIIntegration({
-					recordInputs: true,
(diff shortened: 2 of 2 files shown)


octo-patch:
Addressed in d747589 — lowered the default `tracesSampleRate` to `0.1` (Sentry's recommended production rate) and updated the matching console log. Override path via `options.tracesSampleRate` is preserved, and the changeset entry now mentions the rate drop alongside the privacy defaults.

octo-patch:
Thanks for the follow-up. I think this one is out of scope for this PR and worth tracking separately.

Issue #1681 is specifically about the *Sentry-side* defaults (`sendDefaultPii`, `recordInputs`, `recordOutputs`) — i.e. data Sentry's SDK collects automatically. The `userId`/`briefId` fields here are different: they're explicit application-level correlation IDs that callers pass into `getAITelemetryConfig`, not auto-collected PII. Stripping them changes the contract for every caller and removes the ability to correlate spans with the authenticated user during debugging — even when telemetry is intentionally enabled.

Happy to open a follow-up issue if maintainers want to make `getAITelemetryConfig` honor a "no caller-provided identifiers" flag (e.g. anonymized hashes when an explicit privacy mode is set), but I'd rather keep this PR focused on the Sentry default changes called out in #1681.

PR #1712, opened 2026-06-22
fix(tm-core): slugify tag before using it in file paths (CWE-23 path traversal)

### LINKED ISSUE #1711: Path traversal via unsanitized tag in tm-core file path builders (CWE-23) — regression vs legacy slugify
### Summary

The `tag` value is interpolated **raw** into filesystem paths in two `tm-core`
file-path builders, with no sanitization on the read/generate path. A `tag`
containing `../` escapes the `.taskmaster/` directory, giving a path-traversal
**write** (`*.md`) and **read/probe** (`*.json`) primitive outside the project root.

This looks like a **parity regression**: the legacy JS path slugifies the tag
(`slugifyTagForFilePath`, `scripts/modules/utils.js:348`), but the TypeScript
rewrite of `tm-core` dropped that step. A `sanitizeTag()` helper exists
(`storage.interface.ts:471`) but is **never called** anywhere.

I want to be calibrated up front: this is **not RCE**. The extension is forced
(`.md` on write, `.json` on read) and the written content is task-derived, not
fully attacker-controlled. I'd rate it **Medium** — a real, reachable traversal,
not a critical. Filing as an issue (no SECURITY policy / advisory channel found);
happy to move this to a private advisory if you prefer.

### Affected code (commit `c0c98d3`)

**Write sink** — `packages/tm-core/src/modules/tasks/services/task-file-generator.service.ts:134`
```ts
private getTaskFileName(taskId, tag) {
  return tag === 'master' ? `task_${id}.md` : `task_${id}_${tag}.md`; // tag raw
}
// :93  path.join(outputDir, fileName)  ->  fs.writeFile   (path.join does NOT strip "..")
```

**Read sink** — `packages/tm-core/src/modules/reports/managers/complexity-report-manager.ts:32`
```ts
private getReportPath(tag) {
  const tagSuffix = tag && tag !== 'master' ? `_${tag}` : '';      // tag raw
  return path.join(reportsDir, `task-complexity-report${tagSuffix}.json`);
}
// :55  fs.access + fs.readFile
```

### Why it's reachable (the part that makes it real)

`tag` reaches these sinks **unvalidated** from three sources. `validateTagName`
(the strict `^[a-zA-Z0-9_-]+$` regex in `tag.service.ts`) only runs on
create/rename/copy — never on read/generate.

1. **MCP / CLI arg** — `generate`, `get-tasks`, `get-task`, `set-task-status`
   declare `tag: z.string().optional()` with **no `.regex`/`.refine`**
   (`apps/mcp/src/tools/tasks/generate.tool.ts:22`). The value is passed straight
   to `generateTaskFiles({ tag })`.
2. **`.taskmaster/state.json` → `currentTag`** — read raw in
   `runtime-state-manager.service.ts:53`. This is the zero-interaction vector:
   a repo committed with a poisoned `state.json` triggers traversal as soon as a
   victim runs `task-master generate` / `list` on the clone.
3. **`TASKMASTER_TAG` env var** — same raw assignment (`:63`).

### Reproduction (executed PoC)

I ran this against the **real `tm-core` source** inside an isolated, offline
container (`node:22-alpine`, `--network none`, read-only rootfs). The **read** path
exercises the actual `ComplexityReportManager` class (only its `getLogger` import is
stubbed — all path logic is untouched); the **write** path runs the verbatim
`getTaskFileName` + `path.join` + `fs.writeFile` from the generator. Project root is
`/work/project`; the attacker escapes to `/work/escape` (outside the project).

```
=== READ — real ComplexityReportManager.loadReport(tag) ===
projectRoot         : /work/project
attacker tag        : "../../../../../escape/secret"
RESOLVED read path  : /work/escape/secret.json      <-- outside project
escapes project?    : true
loadReport() result : FILE READ + PARSED (traversal works)
leaked meta.secret  : THIS_FILE_IS_OUTSIDE_THE_PROJECT_ROOT

=== WRITE — verbatim from task-file-generator.service.ts ===
outputDir           : /work/project/.taskmaster/tasks
attacker tag        : "../../../../../escape/pwned"
RESOLVED write path : /work/escape/pwned.md         <-- outside project
escapes project?    : true
write confirmed     : # attacker-written task file (outside project root)
```

Note on depth: the glued prefix (`task-complexity-report_..` / `task_001_..`) is a
literal directory segment, so it costs **two** `..` to undo — five `../` reach the
grandparent of `.taskmaster/` here. An attacker just adds segments; the primitive
is the unsanitized `tag`, not the exact depth.

A poisoned `.taskmaster/state.json` of `{ "currentTag": "../../../../../escape/pwned" }`
reaches the same sink via `generate` with **no flags and no prompt-injection needed**
(`currentTag` is read raw in `runtime-state-manager.service.ts:53`).

### Impact (honest)

- **Write**: create/overwrite any `*.md` file in any writable location
  (e.g. an autostart `.md`, a doc served elsewhere, a README in another repo).
- **Read**: file-existence + `.json`-shape oracle on arbitrary paths (content is
  consumed for complexity enrichment, not returned verbatim — limited exfil).
- Constrained by the forced `.md`/`.json` extension → **not** arbitrary-file
  overwrite of e.g. `~/.bashrc`, and **not** RCE.

CWE-22 / CWE-23 / CWE-73.

### Suggested fix

1. Factor a single **tag-aware path resolver** shared by both builders (kills the
   legacy/TS parity drift), and call `sanitizeTag` (or reject) inside it.
2. Validate `tag` on the **read/generate** path too (same regex as create), or
   reject any tag containing `/`, `\`, `..`, or NUL — including the value read
   from `state.json` and `TASKMASTER_TAG`.
3. Defense-in-depth: assert `resolved.startsWith(baseDir + path.sep)` after resolution.
4. Negative-oracle tests: `../../x`, `..\\..\\x`, NUL → rejected; `feature-v2` → allowed.

I'm happy to send a focused PR if you'd like (I'll follow the repo's CodeRabbit
docstring coverage and AI-assistance disclosure conventions).

---
*Disclosure: this analysis was AI-assisted; findings were human-reviewed and the
PoC above was executed against the real source in an isolated, offline
(`--network none`) container before filing — no network calls, no out-of-sandbox
writes, nothing submitted automatically.*

zied-jlassi:
# What type of PR is this?

 - [x] 🐛 Bug fix

## Description

`tag` is interpolated **raw** into filenames in two `tm-core` path builders:

- `ComplexityReportManager.getReportPath` → `task-complexity-report_<tag>.json` (read)
- `TaskFileGeneratorService.getTaskFileName` → `task_<id>_<tag>.md` (write)

Tags are user/agent/config controllable (CLI `--tag`, MCP `tag` argument,
`.taskmaster/state.json` `currentTag`, `TASKMASTER_TAG`), and none of these entry
points validates the value on the read/generate path (`validateTagName` only runs
on create/rename/copy). So a tag like `../../../../../escape/x` escapes
`.taskmaster/` — an out-of-project file **read** (complexity report) and **write**
(task `.md`). CWE-22/23/73.

This is a parity regression: the legacy JS path slugified the tag
(`slugifyTagForFilePath` in `scripts/modules/utils.js`); the TypeScript rewrite
dropped it, and the `sanitizeTag` helper on the storage interface was never wired in.

**Fix:** a shared `slugifyTagForFilePath` helper (mirroring the legacy behavior),
applied at both path builders and at `cleanupOrphanedFiles` so written names and
the orphan-cleanup pattern stay in sync. Sanitizing at the path-building chokepoint
covers all three entry vectors at once. Legitimately named tags (already
`^[a-zA-Z0-9_-]+$`) are unaffected beyond lowercasing. Honest severity: **Medium** —
the extension is forced (`.md`/`.json`), so this is not arbitrary-file overwrite or RCE.

## Related Issues

Fixes #1711.

## How to Test This

```bash
npm run test -w @tm/core -- tag-path.spec.ts task-file-generator.service.spec.ts
```

**Expected result:** all tests pass, including the new negative-oracle cases where
`../../x`, `..\..\x`, NUL and `..%2f..` are neutralized while `feature-v2` is preserved.

Manual before/after (project root `/work/project`): tag `../../../../../escape/secret`
resolved **outside** the project before the fix and resolves to
`…/.taskmaster/reports/task-complexity-report_escape-secret.json` (contained) after.

## Contributor Checklist

- [x] Created changeset: `npm run changeset`
- [x] Tests pass: `npm test` (vitest 26/26 on the new + touched specs; `tsc --noEmit` clean)
- [x] Format check passes: `npm run format-check` (biome, no changes)
- [ ] Addressed CodeRabbit comments (if any)
- [x] Linked related issues (Fixes #1711)
- [x] Manually tested the changes

## Changelog Entry

Fix path traversal via an unsanitized `tag` in the complexity-report reader and task-file generator (tags are now slugified before use in file paths).

---

*Disclosure: this fix was AI-assisted and human-reviewed; it was tested on the repo's
own vitest/tsc/biome before submission. Happy to follow whatever disclosure format you prefer.*

## Release Notes

* **Bug Fixes**
  * Patched a security vulnerability where crafted tags could be interpolated into report/task filenames, enabling path traversal.
  * Tags are now converted into filesystem-safe, lowercased slugs before being used in filenames, preventing `../`, slashes, and other unsafe characters from escaping the intended directory.
* **Tests**
  * Added coverage to verify safe slug generation, expected fallback behavior, and protection against traversal-style inputs.

### DIFF
--- a/.changeset/fix-tag-path-traversal.md
+++ b/.changeset/fix-tag-path-traversal.md
@@ -1,0 +1,5 @@
+---
+"task-master-ai": patch
+---
+
+Fix path traversal via unsanitized `tag` in tm-core file path builders (CWE-23). The complexity-report reader and the per-task file generator interpolated the tag straight into a filename, so a tag containing `../` (from a CLI/MCP argument, `.taskmaster/state.json` `currentTag`, or `TASKMASTER_TAG`) could read or write files outside the `.taskmaster/` directory. Tags are now slugified with a shared `slugifyTagForFilePath` helper, restoring parity with the legacy JS path writer; legitimately named tags are unaffected.
--- a/packages/tm-core/src/common/utils/index.ts
+++ b/packages/tm-core/src/common/utils/index.ts
@@ -57,6 +57,9 @@ export {
 // Export path construction utilities
 export { getProjectPaths } from './path-helpers.js';
 
+// Export tag-aware path safety utilities
+export { slugifyTagForFilePath } from './tag-path.js';
+
 // Additional utility exports
 
 /**
--- a/packages/tm-core/src/common/utils/tag-path.spec.ts
+++ b/packages/tm-core/src/common/utils/tag-path.spec.ts
@@ -1,0 +1,73 @@
+import { describe, expect, it } from 'vitest';
+import { slugifyTagForFilePath } from './tag-path.js';
+
+describe('slugifyTagForFilePath', () => {
+	describe('legitimate tags are preserved (no mangling)', () => {
+		it('keeps a simple valid tag unchanged', () => {
+			expect(slugifyTagForFilePath('feature-v2')).toBe('feature-v2');
+		});
+
+		it('keeps underscores and digits', () => {
+			expect(slugifyTagForFilePath('release_2026_01')).toBe('release_2026_01');
+		});
+
+		it('lowercases (parity with legacy slugify)', () => {
+			expect(slugifyTagForFilePath('MyFeature')).toBe('myfeature');
+		});
+	});
+
+	describe('path traversal is neutralized (negative oracle)', () => {
+		// These are the inputs an attacker would use to escape the target dir.
+		const traversals = [
+			'../../x',
+			'../../../../../etc/passwd',
+			'..\\..\\x',
+			'..%2f..%2fx',
+			'foo/../bar',
+			'a/b/c',
+			'tag\0null'
+		];
+
+		it.each(traversals)(
+			'produces no path separators or dot segments for %j',
+			(input) => {
+				const slug = slugifyTagForFilePath(input);
+				expect(slug).not.toContain('/');
+				expect(slug).not.toContain('\\');
+				expect(slug).not.toContain('..');
+				expect(slug).not.toContain('\0');
+				// must be a single safe filename fragment
+				expect(slug).toMatch(/^[a-z0-9_-]*$/);
+			}
+		);
+
+		it('cannot escape when embedded in a filename', () => {
+			const slug = slugifyTagForFilePath('../../../../../escape/secret');
+			const fileName = `task-complexity-report_${slug}.json`;
+			expect(fileName).not.toContain('/');
+			expect(fileName).not.toContain('..');
+		});
+	});
+
+	describe('hyphen handling and length', () => {
+		it('collapses and trims hyphens introduced by invalid chars', () => {
+			expect(slugifyTagForFilePath('../../etc/passwd')).toBe('etc-passwd');
+		});
+
+		it('caps the length at 50 characters', () => {
+			const long = 'a'.repeat(120);
+			expect(slugifyTagForFilePath(long).length).toBe(50);
+		});
+	});
+
+	describe('empty / invalid input', () => {
+		it('returns a fallback for empty string', () => {
+			expect(slugifyTagForFilePath('')).toBe('unknown-tag');
+		});
+
+		it('returns a fallback for null/undefined', () => {
+			expect(slugifyTagForFilePath(null)).toBe('unknown-tag');
+			expect(slugifyTagForFilePath(undefined)).toBe('unknown-tag');
+		});
+	});
+});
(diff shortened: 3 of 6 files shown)


zied-jlassi:
Good catch on the empty-slug edge case — a tag that is entirely invalid characters (e.g. `../../../`) collapsed to `""` and would have produced a malformed `task-complexity-report_.json` suffix. Fixed in `cdc59a1`: the helper now returns the same `unknown-tag` sentinel used for empty/non-string input, with tests covering `../../../` and `////`.

